PR #15, opened 2016-02-23
Update README.md

picoworm:
Thank you for your nice library. 
You forgot to change artifact id on mainpage


edvin:
Ah, nice catch! Thank you. I've merged the PR now :)

PR #16, opened 2016-02-23
BorderPane Builder

thomasnield:
I'm trying to implement `BorderPane` builders and get something to this effect.

```kotlin

class BorderPaneTest : App() {

    override val primaryView = BorderPaneItem::class

}

class BorderPaneItem() : View() {
    override val root = BorderPane()
    init {
        with(root) {
            top {
                button("Top") {
                    borderpaneConstraints {
                        alignment = Pos.TOP_CENTER
                    }
                }
            }
            bottom {
                button("Bottom") {
                    borderpaneConstraints {
                        alignment = Pos.TOP_LEFT
                    }
                }
            }
            left {
                button("Left") {
                    borderpaneConstraints {
                        alignment = Pos.TOP_LEFT
                    }
                }
            }
            right {
                button("Right") {
                    borderpaneConstraints {
                        alignment = Pos.BOTTOM_RIGHT
                    }
                }
            }
        }
    }
}
```

I'm getting the "dupe children" issue and I think this has to do with builders adding the elements twice due to a parent inadvertently getting targeted. I think you did some sort of trick with the `GridPane` builder by using some "fake" `Pane`. 

What do you think?

### DIFF
--- a/src/main/java/tornadofx/Builders.kt
+++ b/src/main/java/tornadofx/Builders.kt
@@ -135,44 +135,24 @@ fun Pane.gridpane(op: (GridPane.() -> Unit)? = null) = opcr(this, GridPane(), op
 fun Pane.borderpane(op: (BorderPane.() -> Unit)? = null) = opcr(this,BorderPane(),op)
 
 fun BorderPane.top(op: Pane.() -> Unit) {
-    val pane: Pane = Pane()
-    op(pane)
-    val childCount = pane.children.size
-
-    when(childCount) {
-        1 -> top = pane.children[0]
-        in 1..1000 -> top = HBox().apply{ this.children.setAll(*pane.children.toTypedArray()) }
-    }
+    val vbox = VBox()
+    op(vbox)
+    top = if (vbox.children.size == 1) vbox.children[0] else vbox
 }
 fun BorderPane.bottom(op: Pane.() -> Unit) {
-    val pane: Pane = Pane()
-    op(pane)
-    val childCount = pane.children.size
-
-    when(childCount) {
-        1 -> bottom = pane.children[0]
-        in 1..1000 -> bottom = HBox().apply{ this.children.setAll(*pane.children.toTypedArray()) }
-    }
+    val vbox = VBox()
+    op(vbox)
+    bottom = if (vbox.children.size == 1) vbox.children[0] else vbox
 }
 fun BorderPane.left(op: Pane.() -> Unit) {
-    val pane: Pane = Pane()
-    op(pane)
-    val childCount = pane.children.size
-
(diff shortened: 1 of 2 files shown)


thomasnield:
Oh wait, I think I see what you did there. Clever. Let me try this again.

thomasnield:
Much better! I think I'm getting somewhere. The only limitation is only one node can be passed and the rest will be ignored. We can change that though. Perhaps we can check the count of components provided and handle putting them in a `Pane` if there is more than one.

thomasnield:
Okay I changed that. Just committed my rough proposal for handling multiple `Node` items, default to `HBox` or `VBox`.

thomasnield:
I'll need to simplify and cleanup how multiple children are handled. Will do tomorrow if we move forward with this.

edvin:
If there is only one children, it's important that it is not wrapped in anything. The use should be able to:

```kotlin
top {
    hbox {
        button("Button 1")
        button("Button 2")
    }
}
```

This would keep it in line with how the rest of the builders work. If there are multiple children, we could either throw an exception, or wrap in an hbox or something by default. Unsure what's the best approach, try to experiment a little with it :)

You could use an HBox as the fake, and keep it if there are multiple children, and unwrap if there is a single child maybe?

I'm going to be gone for most of the day, but hopefully I'll be able to look at it tonight. (I'm GMT+1)

edvin:
Hmm.. I think VBox is a better default because of the nature of the BorderPane.

thomasnield:
Pretty much my thoughts too. Will do. And no worries.

thomasnield:
Will get to this later today or tomorrow. But I just wanted to say I used TornadoFX along with RxJava to re-write and enhance a complex UI. It is phenomenal! What took me a week without RxJava and TornadoFX, I did in less than an hour!

edvin:
Great! I feel the same way. I'm much more productive with this stack, and it makes me go the extra mile, so I add features in my apps that earlier felt like a chore to implement, so I might have skipped them :)

thomasnield:
You and I have very similar ambitions and goals : )

edvin:
It seems that way. I'm really glad you got involved in TornadoFX :)

thomasnield:
Me too, I'm happy to contribute. This library definitely shows what Kotlin is capable of, and is probably the most exciting one I found yet.

thomasnield:
I think everything is implemented here. Anything else need to be done?

edvin:
Sweet! I will review and merge it tonight! Still on vacation so things are going a bit slower than usual.

edvin:
Looking good. Nice work, @thomasnield :+1:

Issue #17, opened 2016-02-25
Saving components to properties with builders?


thomasnield:
I hope I don't sound too ignorant. I don't have much practical experience with MVC paradigms, although this library has gotten me very interested in them. I often avoided them in Swing because they felt like they added boilerplate. 

But that aside, I am trying to think about how to save components to properties so I can create and combine RxJava streams off them outside the builders. Is this an acceptable pattern? 

```kotlin
 class MyView: View() {
        override val root = VBox()

        var myButton: Button?

        init {
            with(root) {
                    myButton = button("New Entry")
            }
       }
       fun getEvents(): Observable<ActionEvent> =
            myButton?.toNodeEvents(ActionEvent.Action)
}
```

If so, would it add any value to create a `SingleAssign` property delegate so we can not only rid the null checking, but also ensure a value gets only assigned once?

```kotlin

class SingleAssign<T> {

    @Volatile
    private var initialized = false

    @Volatile
    private var _value: Any? = UNINITIALIZED_VALUE

    operator fun getValue(thisRef: Any?, property: KProperty<*>): T {
        if (!initialized)
            throw Exception("Value has not been assigned yet!")
        return _value as T
    }

    operator fun setValue(thisRef: Any?, property: KProperty<*>, value: T) {
        synchronized(this) {
            if (initialized) {
                throw Exception("Value has already been assigned!")
            }
            _value = value as T
            initialized = true
        }
    }


    private object UNINITIALIZED_VALUE
}
```

Then we could call it like this. 

```kotlin
 class MyView: View() {
        override val root = VBox()

        var myButton: Button by SingleAssign()

        init {
            with(root) {
                    myButton = button("New Entry")
            }
       }
       fun getEvents(): Observable<ActionEvent> =
            myButton.toNodeEvents(ActionEvent.Action)
}
```

edvin:
To get rid of the null check you would declare it as `lateinit var myButton: Button`. That doesn't ensure you don't reassign it though, and `lateinit` is sadly disallowed on `val`. Your solution would prevent that, at the cost of synchronization.

I see the value in this, but I think the most annoying bit is the null check, and that's easily avoided. On the other hand, I can't see any harm in adding it, it's up to the user to use it :)

thomasnield:
I did not know about lateinit modifier and you are right, that solves most of the problem. You could also make the setter private, but that could verbose.  In a multithreaded application, would you still in theory need a volatile modifier on the property too?

[...]

PR #18, opened 2016-02-25
Implement SingleAssign functionality

thomasnield:
Please refer to #17.

Let me know if the synchronization `enum` argument name is too long or if you need unit tests.

```kotlin

class MainApp : App() {
    override val primaryView = MainView::class

    class MainView : View() {
        override val root = GridPane()

        var testVar: String by singleAssign()
        // can also be called with argument singleAssign(SingleAssignThreadSafetyMode.SYNCHRONIZED)
        // or singleAssign(SingleAssignThreadSafetyMode.NONE)
        
        init {

            //CASE 1
            println(testVar) //fails correctly

            //CASE 2
            testVar = "Hello"
            println(testVar)
            testVar = "World" //fails correctly

            //CASE 3
            testVar ="Hello"
            println(testVar) //no issues
        }
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- `singleAssign()` property delegates (https://github.com/edvin/tornadofx/issues/17)
+- Tests for `singleAssign()`
+
+### Added
 - BorderPane builder (https://github.com/edvin/tornadofx/pull/16)
 - `Node.gridpaneConstraints` extension function (https://github.com/edvin/tornadofx/issues/12)
 - `Node.vboxConstraints` extension function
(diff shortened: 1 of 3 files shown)


thomasnield:
Shoot, there's conflicts. Let me try to fix that.

edvin:
It would be great with some simple unit tests as well, the framework is already in place, so maybe you can add just a few simple tests?

thomasnield:
Okay fixed conflicts. I'll add the unit tests now.

edvin:
Sweet! You can include a line in the changelog as well if you'd like :)

edvin:
enum name is fine, and in line with what Kotlin does, so it's all good :)

edvin:
We have so much properties goodness now that we might need a new wiki page for this :)

thomasnield:
Done! And yeah we probably do. Let me know if you need help.

PR #21, opened 2016-02-26
Max height/width constraints for layout builders

thomasnield:
`GridPane`,`VBox`, and `HBox` really need constraint parameters to use the max width, max height, or both. They're not natively part of these layouts but they would be a welcome enhancement to me. What do you think? 

**No useMaxWidth**

**With useMaxWidth**

```kotlin
class BorderPaneItem() : View() {
    override val root = GridPane()
    init {
        with(root) {
            button("Button").gridpaneConstraints {
                columnRowIndex(0,0)
                useMaxWidth = true
            }
            button("Another Button").gridpaneConstraints {
                columnRowIndex(0,1)
                useMaxWidth = true
            }
        }
    }
}
```

### DIFF
--- a/src/main/java/tornadofx/Nodes.kt
+++ b/src/main/java/tornadofx/Nodes.kt
@@ -289,8 +289,16 @@ class GridPaneConstraint(
         var fillHeight: Boolean? = null,
         var fillWidth: Boolean? = null,
         var hAlignment: HPos? = null,
-        var vAlignment: VPos? = null
+        var vAlignment: VPos? = null,
+        var useMaxWidth: Boolean? = null,
+        var useMaxHeight: Boolean? = null
 ) {
+    var useMaxHeightWidth: Boolean? = null
+        set(value) {
+            useMaxWidth = value
+            useMaxHeight = value
+            field = value
+        }
     var vhGrow: Priority? = null
         set(value) {
             vGrow = value
@@ -325,6 +333,8 @@ class GridPaneConstraint(
         fillWidth?.let { GridPane.setFillWidth(node, it) }
         hAlignment?.let { GridPane.setHalignment(node, it) }
         vAlignment?.let { GridPane.setValignment(node, it) }
+        useMaxWidth?.let { if (node is Region && it) node.maxWidth = Double.MAX_VALUE }
+        useMaxHeight?.let { if (node is Region && it) node.maxHeight = Double.MAX_VALUE }
         return node
     }
 }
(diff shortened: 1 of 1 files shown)


edvin:
Shouldn't these be extension functions on Region?

```kotlin
fun <T : Region> T.useMaxWidth(): T {
    maxWidth = Double.MAX_VALUE
    return this
}
```

That way there is no mismatch and you can use it like:

```kotlin
button("Another Button") {
    gridPaneConstraints {
                columnRowIndex(0,1)
    }
    useMaxWidth = true
}
```

Slightly more verbose, but also slightly more correct :) This also gives the added benefit of being available outside the builder. What do you think?

thomasnield:
I was actually figuring that out later last night. Okay. I'll add it as an extension function.

thomasnield:
I'll provide context, because I guess this brings up a larger design question: is this library meant to be an alternative way of expressing JavaFX while keeping the native nodes as they are (with small extension functions here and there)? Or do we want to augment and, in a few cases, re-envision how controls work to make them better?

The reason I proposed this PR is because GridLayout and GridBagLayout in Swing had a similar attribute allowing a control to take up its entire region. When JavaFX removed this attribute from GridPane i felt it was unwarranted. While not natively part of GridPane, I was more or less "putting it back" with this PR.

The extension function works for me. But isn't it a thought to consider adding constraints to GridPane that "ought to be there" even if they technically are not? I guess it comes down to convention vs usefulness...

thomasnield:
and what our balance is between those two factors.

thomasnield:
Then again the more I think about it, you technically are right. It is a Region specifically and it might cause confusion if people try to use the attribute on non-Region Nodes. For that reason I think I will buy into the extension function.

thomasnield:
Sorry for thinking out loud. I'll go ahead and close :)

edvin:
Don't be sorry - it helps to understand the reasoning :) When I think about it, you should actually be able to use it like you want to anyway:

```kotlin
button("Another Button") {
    gridPaneConstraints {
        columnRowIndex(0,1)
        useMaxWidth = true
    }
}
```

`useMaxWidth` will still be applied to the `Button` even though it seems like it is part of the constraints.

Sorry for the late reply, still on vacation so I only get a minute here and a minute there :)

edvin:
Will you add the extension functions to Region?

thomasnield:
Ohhhhh. Forgot about that whole parent-scope thing. Awesome.

And yeah I'll open a new PR.

thomasnield:
I came up with them as extension properties. These good? I'll put in a PR as soon as I'm not behind the corporate firewall. 

```kotlin
var Region.useMaxWidth: Boolean
    get() = maxWidth == Double.MAX_VALUE
    set(value) = if (value) maxWidth = Double.MAX_VALUE else Unit

var Region.useMaxHeight: Boolean
    get() = maxHeight == Double.MAX_VALUE
    set(value) = if (value) maxHeight = Double.MAX_VALUE else Unit

var Region.useMaxHeightWidth: Boolean
    get() = maxWidth == Double.MAX_VALUE && maxHeight == Double.MAX_VALUE
    set(value) = if (value) { useMaxWidth = true; useMaxHeight = true } else Unit
```

edvin:
Looking good!

thomasnield:
I may change the third to `useMaxSize` to follow current conventions in JavaFX...

thomasnield:
Thinking of creating a `usePrefSize` as well, with `usePrefWidth` and `usePrefHeight`.
```
 newEntryButton = button("New Entry") {
                    gridpaneConstraints {
                        columnRowIndex(0,0)
                        margin = Insets(0.0,0.0,5.0,0.0)
                    }
                    useMaxWidth = true
                    setMinSize(Button.USE_PREF_SIZE, Button.USE_PREF_SIZE)
                }
```

edvin:
Yeah, I like `usePrefSize` instead of of `useHeightWidt`. `usePrefSize` sounds good as well.

PR #24, opened 2016-02-27
Max and Preferred Sizing Extension Properties

thomasnield:
See #21


thomasnield:
Darn. Created conflicts with myself with different PR's. This should be fixed now.

Issue #11, opened 2016-02-21
TableView Builder?


thomasnield:
Do you think it is possible to create a `TableView` builder? I think it would go nicely with the `TableView` extensions already present. I can put in a PR and try to work out the details if needed, unless somebody has a faster vision than me on how to do it.

```
tableview {
        items = getMyItems()
        addColumn<ReportItem,String>("Category") { it.value.name }
        addColumn<ReportItem,BigDecimal>("Week") { it.value.weekTotal }
}
```

edvin:
That should be doable :) I'll give it a go during the next week. I'm on vacation though, so I might not get it done before the week after.

thomasnield:
Okay, have a nice vacation! if I have time I may put in a PR just to attempt getting involved in this project. But no worries.

edvin:
Let me know if you wan't to give this a go as well, @thomasnield ? :)

thomasnield:
I think I'm gonna. Maybe tonight, or sometime this week.

thomasnield:
I've got a few ideas. I'm going to review the `TableView` API and plan this carefully over the next few days.

edvin:
Ouch, I just had to give this a crack, actually needed it today. I found a way to avoid repeating the table type argument, so you can write:

```kotlin
tableview<ReportItem> {
    items = getMyItems()
    column<String>("Category") { it.value.name }
    column<BigDecimal>("Week") { it.value.weekTotal }
}
```

It really bothered me to repeat `ReportItem` on every column line. I'm gonna check in my implementation soon, take a look and let me know what you think. Nothing is final, so we can play with it until we get it right :)

`addColumn` is called `column` in the builder, to align it with the rest of the syntax. To avoid repeating `ReportItem` on every column line I had to subclass TableView. This kind of breaks with the rest of the builders, but I think this might be a migration path to add even nicer syntax to builders over time.

What do you think?

thomasnield:
At first glance this looks really good. Kind of what I envisioned. I'll play with it today. And ditto on the column type parameters. Btw do we want to include cell and row rendering factories too?

edvin:
Yes! I already made `cellFormat`: https://github.com/edvin/tornadofx/blob/master/src/main/java/tornadofx/Nodes.kt#L113

edvin:
(If `cellFormat` is too simplistic, we'll change it :)

thomasnield:
[...]

PR #26, opened 2016-02-27
Add TableCell rowItem extension property

thomasnield:
Refer to #11. Would it be okay to add a `rowItem` extension property for this? I think it might be useful especially when using `cellFormat()`.


edvin:
Absolutely!

PR #33, opened 2016-03-02
Some useful functional

picoworm:


thomasnield:
I'll take a look at this tomorrow (Texas time) if Edvin doesn't get to it first.

edvin:
Hi! Thanks a lot for your effort, I really appreciate it. It will take me some time to review everything before I merge, so I expect I'll have some more questions once I get through it all. Some initial questions though:

1. I see there is a new dependency on `kotlin-reflect`. I have specifically tried to avoid this because of the added runtime size, so for earlier uses I've resorted to Java reflection instead of using `kotling-reflect`. What is requiring `kotlin-reflect`, and would it be possible to avoid it with a workaround?

2. Would it be possible to add some very basic tests for the functionality? If you don't have the time, I will give it a try, but please let me know if there are specific areas I should be sure to test.

3. `runFXSync` looks like a nice addition - can you give a use case example I can put in the docs? Have never had the need myself, but I'm sure there are many use cases.

4. I think `TableColumn.makeEditable` should throw an exception if it's not able to make the column editable. What do you think?

Thanks so much for `JsonModelAuto` - been thinking of adding this, you beat me to it :))

edvin:
Cool @thomasnield - Let me know if you have inputs to my questions above as well.

thomasnield:
Didn't look at the code yet, but I do agree with the sentiment on avoiding library dependencies until a critical mass of functionality requires it.

I never really use out-of-the-box concurrency features with JavaFX since I use RxJava with RxJavaFX so heavily. But I'll take a look later since I know many people do.

thomasnield:
And yeah I think makeEditable() would be helpful to throw an error if the ObservableValue is fixed and not editable, like a Binding or ReadOnlyObjectProperty.

thomasnield:
Ah... can't help but look a little. Can we somehow associate a Boolean column with a CheckBox  for the editable column feature rather than simply leaving it as a String?

edvin:
Wow @picoworm - did you manage to avoid using type arguments when you dumped FXTableView? When I tried I was unable to write:

```kotlin
column("Name", Customer::name)
```

I had to write:

```kotlin
column<String>("Name", Customer::name)
```

Can't wait to check this out :)

@thomasnield We can absolutely make an editable CheckBox, I'll add that after I merge :)

edvin:
I know that adding FXTableView was an anti pattern as far as the rest of the builders go, so if we can get rid of it, and still don't have to add type arguments, I'm very happy.

thomasnield:
Me too, I was concerned if FXTableView might cause complications later. Good job if we are able to get rid of it.

picoworm:
Hi guys, thankd s lot for your comments

I will do some tests, examples and bugfixes next week

edvin:
Great @picoworm - looking forward to it!

picoworm:
>>I see there is a new dependency on kotlin-reflect. I have specifically tried to avoid this because of the added runtime size, so for earlier uses I've resorted to Java reflection instead of using kotling-reflect. What is requiring kotlin-reflect, and would it be possible to avoid it with a workaround?

Kotlin reflection is powerful library. Using java reflections on kotlin language is not correct i think.
Kotlin creators splitted their library to stdlib and reflection. Maybe we can do it with tornadofx?

In example, one library without reflection for standart features and second as addon for advanced users with reflection.

>>Would it be possible to add some very basic tests for the functionality? If you don't have the time, I will give it a try, but please let me know if there are specific areas I should be sure to test.

I'll do it later. Now have a lot of main job

>>runFXSync looks like a nice addition - can you give a use case example I can put in the docs? Have never had the need myself, but I'm sure there are many use cases.

Ok

>> I think TableColumn.makeEditable should throw an exception if it's not able to make the column editable. What do you think?

I agree with you. Will do.

picoworm:
Hello everyone, please look at several examples made on tornadofx-samples base
https://github.com/picoworm/tornadofx-samples

edvin:
Thanks @picoworm! I can integrate the examples after we have merged this PR. The tornadofx-samples projects has lots of old syntax, so I'll do a full cleanup next week :)

thomasnield:
From what I've looked at, it looks good. Have we decided on use cases for runFXSync()? I can kind of see why it would be useful. You'd safely update a UI and block until it's done, and then continue the procedure on the current thread (loops or scheduled executor tasks could utilize this). I do everything asynchronously with RxJava so I don't use these sort of patterns, but I can see use cases others would. I vote to include that.

Regarding the need for Kotlin Reflection dependency, I have little experience with reflection but I advise avoiding bringing in additional libraries unless we have a defined path ahead of us that will utilize it heavily, or it will simplify a lot of our current reflection code. I will defer to Edvin on this since he built all the reflection-centric API.

picoworm:
As about runFXSync function it is not my function, I use it about 4 years in my JavaFX software
I have copied it from strackoverflow or following site, I do not remember exactly
http://news.kynosarges.org/2014/05/01/simulating-platform-runandwait/

I do not know why this function is not inside javafx Platform class, but it is very useful in some cases

For example, I have an thread that communicates another computer via tcp-ip in continuous manner, not request-responce as background {} ui {} does.

My usage was next:
Communication thread connects, fills an input buffer, and do runAndWait (runFXSync),
then code runs on FX thread and sets all necessary node properties depends on input buffer data, then fills output buffer and exits, then communication thread resumes and sens output buffer via tcp-ip. Then cycle repeats.

Usage can be very different, so when you need to get a data from FX thread and you are not in fxthread you can run

thread {
            val SomName = arrayOf(String())
            runFXSync {
                SomName[0] = MyTextField.text
            }
            useInCurrentThread(SomName)
}

thomasnield:
Man you need to check out reactive programming and RxJava. It will change your life.

edvin:
I would like to merge this PR soon and make some small modifications after the merge. Are you satisfied with how it looks now @picoworm or do you want to make more changes before I merge?

picoworm:
@thomasnield , thanks a lot for your advice. I will.
@edvin, you can merge

Thank you guys

edvin:
Great! I will merge tomorrow, now it's dinner + movie :)

edvin:
@picoworm I forgot to ask - are you happy with how this PR was integrated? Thanks a lot for your contribution, it was invaluable.

picoworm:
Thanks, I am very glad to share my code here.

PR #37, opened 2016-03-05
Menu Builders

thomasnield:
I have a few more items to add, but I came up with some builders for menus. This look like the perfect candidate for builder patterns.

```kotlin

class MenuApp: App() {
    override val primaryView = MenuView::class
}


class MenuView: View() {

    override val root = VBox()

    init {
        with(root) {
            menubar {
                menu("File") {
                    menu("Switch Account") {
                        menuItem("Facebook") { println("Switching to Facebook") }
                        menuItem("Twitter") { println("Switching to Twitter") }
                    }
                    menuItem("Save") { println("Saving") }
                    menuItem("Exit") { println("Exiting")}
                }
                menu("Edit") {
                    menuItem("Copy") { println("Copying") }
                    menuItem("Paste") { println("Pasting") }
                    menu("Options") {
                        menuItem("Account") { println("Launching Account Options") }
                        menuItem("Security") { println("Launching Security Options") }
                        menuItem("Appearance") { println("Launching Appearance Options") }
                    }
                }
            }
        }
    }
}
```

menus


thomasnield:
Added separators. Trying to figure out what to do with `CheckMenuItem` and `RadioMenuItem` and how to populate values with builders.

```kotlin

class MenuApp: App() {
    override val primaryView = MenuView::class
}


class MenuView: View() {

    override val root = VBox()

    init {
        with(root) {
            menubar {
                menu("File") {
                    menu("Switch Account") {
                        menuItem("Facebook") { println("Switching to Facebook") }
                        menuItem("Twitter") { println("Switching to Twitter") }
                    }
                    separator()
                    menuItem("Save") { println("Saving") }
                    menuItem("Exit") { println("Exiting")}
                }
                menu("Edit") {
                    menuItem("Copy") { println("Copying") }
                    menuItem("Paste") { println("Pasting") }
                    separator()
                    menu("Options") {
                        menuItem("Account") { println("Launching Account Options") }
                        menuItem("Security") { println("Launching Security Options") }
                        menuItem("Appearance") { println("Launching Appearance Options") }
                    }
                }
            }
        }
    }
}
```

menus

thomasnield:
Alright, I'm going to call it for the day. Edvin please let me know what you think when it is daylight for you.

edvin:
Nice! I have some comments:

- I don't think we have talked about it, but all builders have all lowercase letters now, so this seems to be the pattern we use.. should we rename `menuItem` to `menuitem`?

- Defining the action as the lambda parameter breaks with the pattern of other builders, and also makes it a bit more awkward to further customize a MenuItem, let's say for example to add an icon, accelerator etc.

We could add some parameters to the `menuitem` function and keep using the lambda as the action expression:

```kotlin
menuitem("Facebook", Icons.Facebook, KeyCombination.valueOf("Shortcut+F") {
    println("Switching to Facebook")
}
```

Or use the same style as the other builders:

```kotlin
menuitem("Facebook") {
   graphic = Icons.Facebook
   accelerator = KeyCombination.valueOf("Shortcut+F")
   setOnAction { 
      println("Switching to Facebook")
   }
}
```

If we at the same time add the parameters I mention above to menuItem you can call it like this as well:

```kotlin
menuitem("Facebook", Icons.Facebook, KeyCombination.valueOf("Shortcut+F")).setOnAction {
    println("Switching to Facebook")
}
```

I'm not sure what's best, but we at least need a nice way to further customize the menuitem. What do you think?

thomasnield:
Very valid points and kind of why I paused to think before proceeding. Trying to choose one of your proposals though... trying to think through which one.

TornadoFX (as well as Kotlin in general) tries to reduce boilerplate for the most common cases. I think that should be our goal. The common case is to define a text, an action, an accelerator, and maybe an icon. 


Therefore I think arguments for text, an accelerator, an optional graphic, and finally a lambda argument for the action event will be the cleanest for the common cases. If you need more customization than that, you can call apply() afterwards. What do you think?

edvin:
I totally agree, let's go with that!

thomasnield:
Cool, so something like your first proposal.

```kotlin
menuitem("Facebook", KeyCombination.valueOf("Shortcut+F"), Icons.Facebook) {
    println("Switching to Facebook")
}
```

edvin:
That should do it. Now it's time for bed, gotta get up early to see Conor McGregor KO Nate Diaz in the first round :) I'm almost done going over the new functional stuff after the merge, will write some more details about it tomorrow.

thomasnield:
Yeah I'm surprised you are still up! I'll let you crash so you can catch the UFC tomorrow then.

edvin:
Sweet! :) Good night!

thomasnield:
I implemented these modifications, as well as added `ContextMenu` extensions. 

```kotlin
 myNode.addEventHandler(MouseEvent.MOUSE_CLICKED, { me ->
	ContextMenu().apply {
		menu("Stuff") {
			menuitem("One") { println("ONE") }
			menuitem("Two") { println("Two") }
		}
	}.show(this,me.screenX,me.screenY)
})
```

edvin:
Wow, you've been incredibly busy! :)) Will try to catch up with everything this afternoon!

edvin:
Will you add some info to the changelog on the new builders you've added? I'd like to start the release process for 1.3.2 as soon as possible. It's crucial for adoption to get the quickstart archetype out there, and it needs features from 1.3.2 :) We also need to add some documentation, I've almost lost count of all the new features in this release.. hehe :)

thomasnield:
For sure, I'll update the change log as each PR is merged.

thomasnield:
Ohh sorry I wasn't thinking and already merged. Didn't realize you were still going to review. Let me know if you see any issues and I will fix them on the master.

edvin:
No problem, and I don't see any issues, just a quick question. Why did you do:

```kotlin
keyCombination?.let { menuItem.accelerator = it }
```

instead of:

```kotlin
keyCombination?.apply { menuItem.accelerator = this }
```

... by the way? I guess it doesn't matter at all, just wondering because I'm curious :) We're not using the returned value for anything, so `let` seems strange to me, so I just wanted to ask - still learning Kotlin :)

thomasnield:
No actually you are right. It is technically more correct to use apply(). I coalesce and map nullable values so much, it's become a habit.

PR #38, opened 2016-03-06
Listview and Scrollpane builders

thomasnield:
I implemented `ListView` and `ScrollPane` builders. I'll merge when you approve, Edvin.

```kotlin

class MyApp: App() {
    override val primaryView = MenuView::class
}

class MenuView: View() {

    override val root = VBox()

    private val persons = FXCollections.observableArrayList<Person>(
            Person(1,"Samantha Stuart", LocalDate.of(1981,12,4)),
            Person(2,"Tom Marks",LocalDate.of(2001,1,23)),
            Person(3,"Stuart Gills",LocalDate.of(1989,5,23)),
            Person(4,"Nicole Williams",LocalDate.of(1998,8,11))
    )

    init {
        with(root) {
            scrollpane {
                listview(persons) {
                    cellFormat { text = it.name }
                }
            }
            scrollpane {
                tableview<Person> {
                    items = persons
                    column("ID", Person::id)
                    column("Name", Person::name)
                    column("Birthday", Person::birthday)
                    column("Age", Person::age)
                }
            }
        }
    }
}

class Person(id: Int, name: String, birthday: LocalDate) {
    var id by property<Int>()
    fun idProperty() = getProperty(Person::id)

    var name by property<String>()
    fun nameProperty() = getProperty(Person::name)

    var birthday by property<LocalDate>()
    fun birthdayProperty() = getProperty(Person::birthday)

    //assume today is 2016-02-28
    val age: Int get() = Period.between(birthday, LocalDate.now()).years

    init {
        this.id = id
        this.name = name
        this.birthday = birthday
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,6 @@ All notable changes to this project will be documented in this file.
 ### Added
 - `singleAssign()` property delegates (https://github.com/edvin/tornadofx/issues/17)
 - Tests for `singleAssign()`
-
-### Added
 - BorderPane builder (https://github.com/edvin/tornadofx/pull/16)
 - `Node.gridpaneConstraints` extension function (https://github.com/edvin/tornadofx/issues/12)
 - `Node.vboxConstraints` extension function
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
-![Tornado FX Logo](graphics/tornado-fx-logo.png?raw=true "Tornado FX")
-# Tornado FX
+![TornadoFX Logo](graphics/tornado-fx-logo.png?raw=true "TornadoFX")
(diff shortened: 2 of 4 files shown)


thomasnield:
The only possible drawback is I did some fake-pane tactics with `scrollpane`, which sacrifices the last lambda argument that could be used to modify `ScrollPane`. You can always call `apply()` afterwards though.

edvin:
Agree, that's the way to go. Why do you put tableview and listview in a scrollpane, btw? They have their own internal scrollpane :)

thomasnield:
Oh I forgot about that haha. I was going into "Swing mode" all of a sudden.

Issue #47, opened 2016-03-10
TreeTableView Builder


thomasnield:
I'd like to explore streamlining the creation of `TreeTableView` with a builder in a similar manner as `TableView`. Not quite sure of the details yet but I imagine the nested structure nature would make it a a good candidate.

edvin:
Nice! I'll actually need TreeTableView next week, so this is perfect timing. I don't know the API of it well enough yet, but I'll start reading up on it in the meantime.

thomasnield:
Yeah me too, although I explored it a little a few months ago. Depending on the complexity this may become a joint effort.

thomasnield:
We should probably do both `TreeView` and `TreeTableView`. I need to think about how to tackle this though.

https://docs.oracle.com/javafx/2/ui_controls/tree-view.htm
https://docs.oracle.com/javase/8/javafx/user-interface-tutorial/tree-table-view.htm

thomasnield:
Man trying to figure out a design for just `TreeView` is challenging enough. Mind is spinning but I know there is a way to do this...

thomasnield:
Trying to visualize a `TreeView` in its usage. 

Have not used `TreeView` and `TreeTableView` that much, but I know they are incredibly painful to build because of the imperative grouping process. Maybe we can set an `ObservableList` as the backing list and functionally group up and nest items somehow. For example, we can group up `employees` by `department`, and then show each indvidual. Speaking out loud here...

```kotlin
val employees: ObservableList<Employee> = ...

treeview<Employee>(employees) { 
	treeitem<String>(Employee::department) {
		treeitem<String>(Employee::name) {
		}
	}
}
```

thomasnield:
Okay, I got something. 

I had to break a few rules because `TreeView` is [not streamlined at all from  a coding perspective](https://docs.oracle.com/javafx/2/ui_controls/tree-view.htm). I had to create extensions of `TreeView`  and `TreeItem` to give the notion of a "backing list" rather than looping and adding raw values. Using functional approaches, we can add functionality that satisfies common use cases for `TreeView`. 

It is still rough but here is a proof-of-concept. Once we leverage some property mapping with reflection, we could have a pretty strong API.

```kotlin

class MyView : View() {

    override val root = VBox()

    init {
        val persons = FXCollections.observableArrayList(
                Person("Mary Hanes","Marketing"),
                Person("Steve Folley","Customer Service"),
                Person("John Ramsy","IT Help Desk"),
[...]

PR #50, opened 2016-03-12
First pass at API for TreeView

thomasnield:
See #47.


thomasnield:
Going with @edvin's implementation as discussed in #47

PR #52, opened 2016-03-18
Fix makeEditable for S == Boolean

Kiskae:
The previous version simply did not work when given a kotlin.Boolean type


edvin:
Thank you! I have added a test to verify that it's working now as well :) I will release a version next where this fix will be present.

PR #53, opened 2016-03-22
Make the gpg signing only happen when mvn deploy is being done

ronsmits:
I needed the `by fxml(filename)` functionality. Doing mvn install kicks of the gpg signing. Usually this is during the deploy phase


edvin:
Thanks @ronsmits - that's a better default behavior anyway :) Merged now. I'm working on a few features and will do a release next week as well.

PR #61, opened 2016-03-29
Tests for Wiki Examples.

hastebrot:
_Work in progress. Do not pull._

This adds a simple test for wiki builder example 1. It loads the example into the primary stage and captures a screenshot of the root node.

The plan is to add a test resource file and compare it visually with the actual screenshot. The capture API is planned to be stabilized in the next TestFX release.


edvin:
Wow, that's pretty sweet! This opens up a lot of interesting test cases, and possibly other applications as well. Nicely done!

hastebrot:
I think screenshot tests are great, given that the builders are used for fairly complex layouts. I only hope the resulting images are more or less identical between Linux, Mac and Windows.

hastebrot:
When I change `useMaxWidth` within one of the `textfield`s, I get a difference of 32 pixels. This is below the default error threshold, and causes the tests not to fail. Some fine-tuning of the pixel difference thresholds is needed, maybe it can be set to 0 (no different colors for pixels allowed).

When I change one of the `Insets` of `hboxContraints` is get a out-of-bounds exceptions for the two image the test compares, due to different image sizes. This needs to be fixed by allowing different sizes.

hastebrot:
Oh, it's just the cursor, that causes the difference (red color). Seems to be a simple capture timing problem.

example-1-difference

edvin:
Ah! Maybe we can hide the cursor?

-- Edvin

hastebrot:
>hide the cursor

There is `primaryStage.scene.focusOwner` which could help. Or this CSS rule `.text-input { -fx-display-caret: false; }`.

edvin:
Maybe the best option is to add a css file to the root for all tests. There we can remove the caret, and adjust other things if needed down the road?

hastebrot:
This is a good idea. It allows us to verify the focus state. In the last commit I used the third suggestion (`root.requestFocus()`), which removes the focus state.

edvin:
Nice! Will you add it to this PR?

hastebrot:
Yes, I'll add something like `test-stylesheet.css` with `.text-input { -fx-display-caret: false; }` to the PR.

Examples are getting more and more (https://github.com/edvin/tornadofx/wiki/Type-Safe-Builders), but I'm sure, when I've finished TestFX's image capture/matcher API, implementing the tests for the examples will be a lot of fun.

edvin:
Oh nice! I'll help out with adding tests once the functionality/infrastructure is in place :))

edvin:
Closing this, open it if you find the time to finish it later, OK? :)

PR #63, opened 2016-03-30
makeIndexColumn for TableView

picoworm:


edvin:
Nice! Maybe you could make the tablecolumn name property configurable as well? Something like:

```kotlin
makeIndexColumn(name: String = "#", startNumber: Int = 1)
```

Why did you change `TableColumn.CellDataFeatures<S, T>` to `TableColumn.CellDataFeatures<S?, T?>`? I think I remember we changed something similar to this earlier, but I'm not sure why.

thomasnield:
So this is a factory to make a column whose header is its index number? Did i interpret this correctly? Interesting...

picoworm:
Using @hastebrot  code I make an test with snapshot of new function
example-table

@edvin , please do not merge TableColumn.CellDataFeatures<S, T> change, Commiting it it was my mistake, sorry

edvin:
OK, I understand :) Hmm... I don't think it's possible to apply only parts of a PR. Could you maybe copy your changes and do a git pull and apply your changes to HEAD and create a new PR? If not, I could always copy out your changes and apply them to HEAD and just close the PR, but it would be nicer to just accept a PR from you :) Can you do that?

PR #64, opened 2016-03-31
added makeIndexColumn for tableview #2

picoworm:


edvin:
Thank you! I lowercased the `Items` parameter name and only applied the items if they were not null :)

PR #65, opened 2016-04-01
JsonModelAuto bugfixes and enableTextWrap() for tablecolumn feature

picoworm:


edvin:
Thank you :) I'll merge and then change enableTextWrap so it returns the column for easy chaining :)

PR #73, opened 2016-04-07
Adding Accordion Support

t-boom:
I'm working on functionality to add Accordions. This is as far as I've got, but I can't think of a clean way to modify the "folds" (such as adding a title, changing the padding, etc.).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file.
 - New `TableColumn` cell factories: `makeComboBox`, `makeTextField`, `makeChoiceBox`, `makeProgressBar`, `makeCheckbox` and `makeDatePicker` (https://github.com/edvin/tornadofx/issues/67)
 - `TableColumn.enableTextWrap` (https://github.com/edvin/tornadofx/pull/65)
 - `splitpane` builder (https://github.com/edvin/tornadofx/issues/72)
+- `accordion` builder (https://github.com/edvin/tornadofx/pull/73)
 
 ### Changed
 
--- a/src/main/java/tornadofx/Builders.kt
+++ b/src/main/java/tornadofx/Builders.kt
@@ -157,10 +157,11 @@ fun Pane.accordion(vararg panes: TitledPane, op: (Accordion.() -> Unit)? = null)
     return accordion
 }
 
-fun Accordion.folds(op: (Pane.() -> Unit)) {
-    val fake = Pane()
-    op(fake)
-    panes.setAll(fake.children.map { TitledPane(null, it) })
+fun Accordion.fold(title: String? = null, op: (Pane.() -> Unit)) {
+    val vbox = VBox()
+    op(vbox)
+    val fold = TitledPane(title, if (vbox.children.size == 1) vbox.children[0] else vbox)
+    panes += fold
 }
 
 fun Pane.hbox(spacing: Double? = null, children: Iterable<Node>? = null, op: (HBox.() -> Unit)? = null): HBox {
@@ -248,7 +249,7 @@ inline fun <S> Pane.treetableview(root: TreeItem<S>? = null, op: (TreeTableView<
     return treetableview
 }
 
-fun <S> TableView<S>.makeIndexColumn(name: String = "#", startNumber:Int = 1): TableColumn<S, Number> {
+fun <S> TableView<S>.makeIndexColumn(name: String = "#", startNumber: Int = 1): TableColumn<S, Number> {
     return TableColumn<S, Number>(name).apply {
         isSortable = false
         prefWidth = width
--- a/src/main/java/tornadofx/Builders.kt
+++ b/src/main/java/tornadofx/Builders.kt
@@ -149,6 +149,20 @@ fun SplitPane.items(op: (Pane.() -> Unit)) {
     items.addAll(fake.children)
 }
 
+fun Pane.accordion(vararg panes: TitledPane, op: (Accordion.() -> Unit)? = null): Accordion {
+    var accordion = Accordion()
+    if (panes.isNotEmpty())
+        accordion.panes.addAll(panes)
+    opcr(this, accordion, op)
+    return accordion
+}
+
+fun Accordion.folds(op: (Pane.() -> Unit)) {
+    val fake = Pane()
+    op(fake)
+    panes.setAll(fake.children.map { TitledPane(null, it) })
+}
+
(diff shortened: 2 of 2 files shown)


t-boom:
I had another idea where I just added a `Accordion.fold(String?, () -> TitledPane` method, but that would have required adding a bunch of methods to create a node the `TitlePane` content. That may end up being a better way to go.

t-boom:
Here's a demo:

```kotlin
class AccordionTest : View() {
    override val root = StackPane()

    init {
        with(root) {
            prefHeight = 400.0
            prefWidth = 400.0

            accordion {
                folds {
                    textarea { promptText = "Notes" }
                    label("Test 2")
                    label("Test 3")
                }
            }
        }
    }
}
```

accordion-test

edvin:
Thanks for this :)

Some questions :) Why did you call the method `folds`? Is that a name common for accordions? It also seems that you can

edvin:
Forget my question about the function name, folds seems to be a good choice/common name for accordions :)

t-boom:
I can change the name, it just popped into my head and I went with it :)

As far as the fold method goes, I mentioned that in the second comment to this PR. The problem is, `TitledPane` is not a `Pane`, so you would have to create a large number of methods for `TitledPane`. That's not a big deal, but `TitledPane` only takes a single node as content. You can easily do that in your example with something like:

```kotlin
accordion {
    fold("Title here") {
        content = Label("Cool Stuff")
    }
}
```

but then (unless I'm missing something) you can't continue to use the TornadoFX DSL for its content.

t-boom:
(I just started messing around with Kotlin and TornadoFX this week, so I may easily have missed something. Sorry :) )

edvin:
Ah, sorry, it's a bit late here and I'm off my game :)

After you have collected the children into a fake VBox, check if there is only one child, and if so add it as the node for the TitledPane. If there are multiple children in the VBox, just add the vbox as the content node of the TitledPane.

edvin:
```kotlin
accordion {
    fold("Tab 1") {
        label("I will be added as the TitledPane node")
    }
    fold("Tab 2") {
        label("I will be added to the VBox and the VBOx will be the TitledPane node")
        label("Me too!")
    }
}
```

edvin:
The same pattern is used in the `scrollpane` builder:

```kotlin
fun Pane.scrollpane(op: (Pane.() -> Unit)) {
    val vbox = VBox()
    op(vbox)
    val scrollPane = ScrollPane()
    scrollPane.content = if (vbox.children.size == 1) vbox.children[0] else vbox
    this += scrollPane
}
```

t-boom:
I'll do that.

Should I worry about some sort of method to change the pane used? Like an extension property that could be set to `VERTICAL`, `HORIZONTAL`, or `STACKED` (that would resolve to a `VBox`, `HBox`, or `StackedPane` respectively?

I'm not sure how hard that would be to implement, but it popped into my head when I was trying to come up with solutions.

edvin:
That's not really necessary, because if you want an HBox instead, simply do:

 accordion {
    fold("Tab 1") {
        hbox {
            label("Look ma, I'm in an HBox!")
            label("Me too!")
        }
    }
}

t-boom:
Fair point, I didn't think about that :)

edvin:
Sweet! Make the changes and I'll happily accept the PR :)

edvin:
If you want you can also add an entry to the CHANGELOG with reference to this issue.

t-boom:
I've added those changes, but there are still some shortcomings. You can't change any settings of the `TitledPane` without saving it off to a variable (such as padding). Is that acceptable?

thomasnield:
I haven't looked at it since I'm on my phone. But if the TitledPane is returned by the builder, you can always call apply() afterwards to make those manipulations in a block.

t-boom:
(Compulsory example :) )

```kotlin
class AccordionTest : View() {
    override val root = StackPane()

    init {
        with(root) {
            prefHeight = 400.0
            prefWidth = 400.0

            accordion {
                fold("Tab 1") {
                    label("I will be added as the TitledPane node")
                }
                fold("Settings") {
                    label("Name")
                    textfield { promptText = "Enter a name" }
                    label("Name 2")
                    textfield { promptText = "Enter another name" }
                }
            }
        }
    }
}
```

accordion-test-2

edvin:
You could add another parameter that is a function that operates on the TitledPane. This would have to be added before the last op, something like:

```kotlin
fun Accordion.fold(title: String? = null, titledPaneOp: (TitledPane.() -> Unit)? = null, op: (Pane.() -> Unit)) {
    val vbox = VBox()
    op(vbox)
    val fold = TitledPane(title, if (vbox.children.size == 1) vbox.children[0] else vbox).apply {
        titledPaneOp?.invoke(this)
    }
    panes += fold
}
```

It won't be as beautiful when you need it though:

```kotlin
fold("Hello", { padding = Insets.EMPTY }) {
    label("Test")
}
```

EDIT: Just saw @thomasnield 's comment. That might be a better idea, not sure which of the alternatives I like best. What about you guys?

edvin:
Hmm.. I think Thomas is right. If you return the TitledPane from the fold function the code becomes:

```kotlin
fold("Hello") {
    label("Test")
}.apply {
    padding = Insets.EMPTY
}
```

t-boom:
I'm leaning towards Thomas's idea as well. I tried it and it seems to work fine.

edvin:
I gotta get to bed, but if you make that last change and return the TitledPane, I'm sure Thomas can merge this PR for you :))

thomasnield:
Sounds good, let me know when you want me to merge and ill look.

t-boom:
Goodnight :)

@thomasnield Using your approach works great. For example:

```kotlin
fold("Settings") {
    label("Name")
    textfield { promptText = "Enter a name" }
    label("Name 2")
    textfield { promptText = "Enter another name" }
}.apply {
    this@accordion.expandedPane = this
}
```

makes the Settings fold expanded. I'll push it up.

t-boom:
It's pushed. Check it out when you get the change. I'll be off for the night in about 30 min, so if you need more changes after that, they'll have to wait till tomorrow :)

thomasnield:
Merged. Looks good to me. Thanks for putting this together and I'm glad you got involved so quickly!

thomasnield:
I do think it is good practice for control builders to return the control anyway, not just so apply() can be called for specialized cases like this but also to be able to save the controls to variables.

edvin:
Absolutely, I'll make sure they all do right away :) That was the initial intent and almost every builder does that, but I found a couple of stragglers! Good catch, Thomas!

PR #75, opened 2016-04-08
Created a TableView.column extension that uses a PropertyValueFactory…

ronsmits:
Created a TableView.column extension that uses a PropertyValueFactory as provider. This way you can easily hook up POJO's.

It can be used like this:
````
override val root = TableView<CategoryStory>()

    init {
        with(root) {
            column("Title", PropertyValueFactory<CategoryStory, String>("name"))
            column("Types", PropertyValueFactory<CategoryStory, String>("types"))
        }
    }
````

### DIFF
--- a/src/main/java/tornadofx/Builders.kt
+++ b/src/main/java/tornadofx/Builders.kt
@@ -284,6 +284,16 @@ fun <S, T> TableView<S>.column(title: String, valueProvider: (TableColumn.CellDa
     return column
 }
 
+/**
+ * Create a column with a PropertyValueFactory as valueProvider
+ */
+fun <S, T> TableView<S>.column(title: String, valueProvider: PropertyValueFactory<S, T>): TableColumn<S, T> {
+    val column = TableColumn<S, T>(title)
+    column.cellValueFactory = valueProvider
+    columns.add(column)
+    return column
+}
+
 fun <S, T> TableColumn<S, T?>.makeComboBox(items: ObservableList<T>, afterCommit: ((TableColumn.CellEditEvent<S, T?>) -> Unit)? = null): TableColumn<S, T?> {
     cellFactory = ComboBoxTableCell.forTableColumn(items)
     setOnEditCommit {
--- a/src/main/java/tornadofx/Builders.kt
+++ b/src/main/java/tornadofx/Builders.kt
@@ -285,11 +285,11 @@ fun <S, T> TableView<S>.column(title: String, valueProvider: (TableColumn.CellDa
 }
 
 /**
- * Create a column with a PropertyValueFactory as valueProvider
+ * Create a column using the propertyName of the attribute you want shown.
  */
-fun <S, T> TableView<S>.column(title: String, valueProvider: PropertyValueFactory<S, T>): TableColumn<S, T> {
+fun <S, T> TableView<S>.column(title: String, propertyName: String): TableColumn<S, T> {
     val column = TableColumn<S, T>(title)
-    column.cellValueFactory = valueProvider
+    column.cellValueFactory = PropertyValueFactory<S, T>(propertyName)
     columns.add(column)
     return column
 }

### CODE REVIEW

**Comment by thomasnield** [comment]
I can see what you're trying to achieve here by supporting the conventional JavaFX `PropertyValueFactory`, but what's wrong with the existing column extension factories? They basically do the same thing but more succinctly...

```kotlin
override val root = TableView<CategoryStory>()

    init {
        with(root) {
            column("Title",CategoryStory::name))
            column("Types", CategoryStory::types)
        }
    }
```

thomasnield:
It also supports POJO's and Property conventions alike... including both read and write operations.

ronsmits:
Hmmz

That works for a class made with kotlin (CategoryStory) it does not work with a 'normal java class'. CategoryStory is a kotlin data class, Chapter is a normal java pojo the error message is 

````
[ERROR] /Users/ronsmits/src/dl/javafx/src/main/java/org/ronsmits/dl/ui/CategoriesList.kt: (33, 13) None of the following functions can be called with the arguments supplied: 
public fun <S, T> javafx.scene.control.TableView<org.ronsmits.dl.Chapter>.column(title: kotlin.String, valueProvider: (javafx.scene.control.TableColumn.CellDataFeatures<org.ronsmits.dl.Chapter, ???>) -> javafx.beans.value.ObservableValue<???>): javafx.scene.control.TableColumn<org.ronsmits.dl.Chapter, ???> defined in tornadofx
public inline fun <S, reified T> javafx.scene.control.TableView<org.ronsmits.dl.Chapter>.column(title: kotlin.String, observableFn: kotlin.reflect.KFunction<javafx.beans.value.ObservableValue<???>>): javafx.scene.control.TableColumn<org.ronsmits.dl.Chapter, ???> defined in tornadofx
public fun <S, T> javafx.scene.control.TableView<org.ronsmits.dl.Chapter>.column(title: kotlin.String, prop: kotlin.reflect.KMutableProperty1<org.ronsmits.dl.Chapter, ???>): javafx.scene.control.TableColumn<org.ronsmits.dl.Chapter, ???> defined in tornadofx
public fun <S, T> javafx.scene.control.TableView<org.ronsmits.dl.Chapter>.column(title: kotlin.String, prop: kotlin.reflect.KProperty1<org.ronsmits.dl.Chapter, ???>): javafx.scene.control.TableColumn<org.ronsmits.dl.Chapter, ???> defined in tornadofx
@kotlin.jvm.JvmName public fun <S, T> javafx.scene.control.TableView<org.ronsmits.dl.Chapter>.column(title: kotlin.String, prop: kotlin.reflect.KProperty1<org.ronsmits.dl.Chapter, javafx.beans.value.ObservableValue<???>>): javafx.scene.control.TableColumn<org.ronsmits.dl.Chapter, ???> defined in tornadofx
````

**Comment by thomasnield** [comment]
Interesting. I'll take a look tomorrow although @edvin is a better authority on this piece.

**Comment by edvin** [comment]
Ron is absolutely right. The best we could do for POJO's with the current pattern would be to support readonly properties via the getter, like this:

```kotlin
column("Title",CategoryStory::getName)
```

To make it writable with the current approach, one would have to do:

```kotlin
column("Title",CategoryStory::getName, CategoryStore::setName)
```

This has the advantage of beeing type safe, but possibly slightly more verbose.

I think the `PropertyValueFactory` approach is nicer, I'll just have a quick look and see if we can make it a bit less verbose :)

edvin:
With a slight change, we can boil it down to this:

```kotlin
column<CategoryStory, String>("Title", "name")
```

I think that looks cleaner than:

```kotlin
column("Title", PropertyValueFactory<CategoryStory, String>("name"))
```

It also hides the fact that we use a PropertyValueFactory under the hood, which makes us able to change the underlying implementation if we so choose, without changing the API.

The implementation would be something like:

```kotlin
fun <S, T> TableView<S>.column(title: String, propertyName: String): TableColumn<S, T> {
    val column = TableColumn<S, T>(title)
    column.cellValueFactory = PropertyValueFactory<S, T>(propertyName)
    columns.add(column)
    return column
}
```

What do you think, Ron? If you agree, change the PR and I'll happily accept it. We might accept it as it is now as well, if you feel that your approach is better - maybe I'm missing something :)

ronsmits:
I am in a workshop today till tomorrow end of day, I'll make a small test thingy in the weekend and then change the PR

edvin:
Great, thanks!

ronsmits:
The change as described by edvin works nice. I updated my simple [example repository](https://github.com/ronsmits/example-tornadofx) to use it. 
In that example I also used this function:
````
fun <S, T> TableView<S>.column(propertyName: String): TableColumn<S, T> {
    val column = TableColumn<S, T>(propertyName)
    column.cellValueFactory = PropertyValueFactory<S, T>(propertyName)
    columns.add(column)
    return column
}
````

Dont know if this is not too generic

edvin:
Very nice, Ron! This will be a fine addition to the TableView support in TornadoFX :) Hereby merged :)

PR #77, opened 2016-04-08
JsonStructure.toPrettyString added

picoworm:
Hello, guys.

How about this function inside tornadofx?


edvin:
Nice! With this, we could even add a property to the Rest API so that the JSON sent from the client can be pretty printed if the user wants it. Are you satisfied with it so I can merge?

picoworm:
You can merge

edvin:
Thanks a lot! I will add it to the CHANGELOG as well.

PR #105, opened 2016-05-02
Added a function to add a separator to a contextMenu

ronsmits:
Its quite normal to be able to add a separator to a context menu
screenshot 2016-05-02 21 35 10

### DIFF
--- a/src/main/java/tornadofx/Menu.kt
+++ b/src/main/java/tornadofx/Menu.kt
@@ -49,14 +49,14 @@ fun ContextMenu.menuitem(name: String, keyCombination: KeyCombination? = null, g
 /**
  * Add a separator to the contextmenu
  */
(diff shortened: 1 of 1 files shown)


src/main/java/tornadofx/Menu.kt
```diff
@@ -46,6 +46,13 @@ fun ContextMenu.menuitem(name: String, keyCombination: KeyCombination? = null, g
     return menuItem
 }
 
+/**
+ * Add a separator to the contextmenu
+ */
+fun ContextMenu.separator(){
```
t-boom: You might want to have this accept a `SeparatorMenuItem.() -> Unit`

thomasnield: I must have missed that. I'll take a closer look later but feel free to make the edit if you want.

ronsmits:
sorry for the mess up, I should not do this fast and late

PR #109, opened 2016-05-06
Feature/css cleanup

t-boom:
Someone should check this to make sure I didn't break anything. I checked it, and it should be good, but it's always nice to have a second pair of eyes.


edvin:
It looks good! Thanks a lot for doing this.

PR #122, opened 2016-06-10
ViewModel / Validation (prev. ModelWrapper)

johnp:
PR for easier code review/design discussion.

There's quite a bit of outdated things in there and things I wasn't sure if it makes sense to keep them. I just didn't want to trash something that may turn out to be actually useful.
You can probable skip looking at all the PropertyField implementations, because I think the GenericPropertyField (gonna rename that) is the way forward, so at the end we'd have PropertyFieldImpl, ListPropertyFieldImpl and SetPropertyFieldImpl. 
I've overloaded the String-`field()` methods quite a lot with several slightly different versions and explained the reasoning behind them. Those that are worth keeping would then also be added to the Number/Object/List/Set `field()` methods.

At the end of the file there are several extension methods and ViewModel classes that are just things I tried out and that still need to be reworked somewhat. ViewSingleModel, although badly named, is probably the most useful thing there.

I'm quite confident that `isDifferent` and `isDirty` ReadOnlyBooleanProperties are worth implementing, but a lot less sure whether or not `validator`/`isValid` is a good idea.

Next up I'd unify the PropertyField implementations as explained above, implement the ReadOnlyBooleanProperties and generally rename/clean-up things.

Comments/Suggestions are greatly appreciated. As this is refactored from the Java implementation and no clean-room design by an experienced Kotlin developer, there may be some things that could be done differently and, if you have doubts regarding the overall design, we can always start from a clean slate.

### DIFF
--- a/src/main/java/tornadofx/ModelWrapper.kt
+++ b/src/main/java/tornadofx/ModelWrapper.kt
@@ -175,7 +175,8 @@ class ModelWrapper<M : Any>
         set(value) = modelProperty.set(value)
 
     /**
-     * This interface defines the operations that are possible for each field of a wrapped class.
+     * This abstract class defines the operations that are possible for each field of a wrapped class and implements
+     * [equals] and [hashCode] based on the field name.
      *
      * @param T
      * *            target type. The base type of the returned property, f.e. [String].
@@ -186,16 +187,16 @@ class ModelWrapper<M : Any>
      * @param R
      * *            return type. The type of the Property that is returned via [property], e.g. [StringProperty].
      */
-    internal interface PropertyField<T, in M, out R : Property<T>> {
-        fun commit(wrappedObject: M)
+    internal abstract class PropertyField<T, in M, out R : Property<T>> {
+        abstract fun commit(wrappedObject: M)
 
-        fun reload(wrappedObject: M)
+        abstract fun reload(wrappedObject: M)
 
-        fun resetToDefault()
+        abstract  fun resetToDefault()
 
-        fun updateDefault(wrappedObject: M)
+        abstract fun updateDefault(wrappedObject: M)
 
-        val property: R
+        abstract  val property: R
 
         /**
          * Determines if the value in the model object and the property field are different or not.
@@ -208,7 +209,7 @@ class ModelWrapper<M : Any>
          * @return `false` if both the wrapped model object and the property field have the same value,
          * *         otherwise `true`
          */
-        fun isDifferent(wrappedObject: M): Boolean
+        abstract fun isDifferent(wrappedObject: M): Boolean
 
         // TODO: It's probably better to fully do this on the ViewModel level instead and have an overridable method stub
         //       just like [ViewModel.commit] there. Advantage: Access to multiple field at the same time allows
@@ -221,32 +222,44 @@ class ModelWrapper<M : Any>
         // those without iterating through all fields; depends on whether or not we want to expose differentFields/dirtyFields
         // as an ObservableList; otoh it would also be nice to have something like Property<*>.differentProperty, maybe
         // as an extension function (see @bottom))
-        fun isValid(): Boolean
+        abstract fun isValid(): Boolean
 
         //fun differentProperty(): ReadOnlyBooleanProperty
         //fun dirtyProperty(): ReadOnlyBooleanProperty
+
+        @Suppress("UNCHECKED_CAST")
+        override fun equals(other: Any?): Boolean {
+            if (this === other) return true
+            if (other?.javaClass != javaClass) return false
+
+            other as PropertyField<*, *, *>
+            return property.name == other.property.name
+        }
+
+        override fun hashCode(): Int {
+            return property.name.hashCode()
+        }
     }
 
     /**
-     * So, this could easily replace KPropertyField and FxPropertyField.
-     * The same can probably be done with KListPropertyField/FxListPropertyField and the Set-equivalents
-     * Could also name this PropertyFieldImpl or sth else
+     * An implementation of [PropertyField] that is used when the fields of the model class are primitives,
+     * JavaFX properties or simple objects.
      */
-    private inner class GenericPropertyField<T, out R : Property<T>>
-    constructor(
-            name: String,
-            private val getter: (M) -> T,
-            private val setter: ((M, T) -> Unit)?,
-            private var defaultValue: T? = null,
-            propertySupplier: (M?, String?) -> R,
-            private val validator: (T) -> Boolean = { t: T -> true }
-               ) : PropertyField<T, M, R> {
+    private inner class PropertyFieldImpl<T, out R : Property<T>>
+    constructor(name: String,
+                private val getter: (M) -> T,
+                private val setter: ((M, T) -> Unit)?,
+                private var defaultValue: T? = null,
+                propertySupplier: (M?, String?) -> R,
+                private val validator: (T) -> Boolean = { t: T -> true }
+               ) : PropertyField<T, M, R>() {
 
         constructor(accessor: KProperty1<M, T>,
                     defaultValue: T? = null,
                     propertySupplier: (M?, String?) -> R,
                     validator: (T) -> Boolean = { t: T -> true }
-                   ) : this(accessor.name, accessor.getter, if (accessor is KMutableProperty1<M, T>) accessor.setter else null,
+                   ) : this(accessor.name, accessor.getter,
+                if (accessor is KMutableProperty1<M, T>) accessor.setter else null,
                 defaultValue, propertySupplier, validator)
 
         //@JvmName("FxPropertyField")
@@ -303,248 +316,48 @@ class ModelWrapper<M : Any>
         override fun isValid(): Boolean {
             return validator(property.value)
         }
-
-        @Suppress("UNCHECKED_CAST")
-        override fun equals(other: Any?): Boolean {
-            if (this === other) return true
-            if (other?.javaClass != javaClass) return false
-
-            other as GenericPropertyField<*, *>
-            return property.name == other.property.name
-        }
-
-        override fun hashCode(): Int {
-            return property.name.hashCode()
-        }
     }
 
     /**
-     * An implementation of [PropertyField] that is used when the fields of the model class are JavaFX Properties too.
-     * TODO: Should we accept ReadOnly(List|Set)?Property for the Fx(Set|List)?PropertyFields and gracefully ignore commit()s
-     *       or throw an exception; and what about KProperty vs KMutableProperty?
+     * An implementation of [PropertyField] that is used when the field of the model class is a [List] or a JavaFX [ListProperty].
+
+     * @param E
+     * *            the type of the list elements.
      */
-    private inner class FxPropertyField<T, R : Property<T>>
+    private inner class ListPropertyFieldImpl<E, T : ObservableList<E>, out R : Property<T>>
     //@JvmOverloads // causes CompilationException!
     constructor(name: String,
-                private val accessor: (M) -> Property<T>,
-                private var defaultValue: T? = null,
-                propertySupplier: (M?, String?) -> R,
-                private val validator: (T) -> Boolean = { t: T -> true }
-               ) : PropertyField<T, M, R> {
-
-        constructor(accessor: KProperty1<M, Property<T>>,
-                    defaultValue: T? = null,
-                    propertySupplier: (M?, String?) -> R,
-                    validator: (T) -> Boolean = { t: T -> true }) : this(accessor.name, accessor, defaultValue,
-                propertySupplier, validator)
-
-        override val property: R // TODO: use property.name for equals()/hashCode() now that it's set?
-
-        init {
-            this.property = propertySupplier(null, name)
-            this.property.addListener { observable, oldValue, newValue -> this@ModelWrapper.propertyWasChanged() }
-        }
-
-        override fun commit(wrappedObject: M) {
-            accessor(wrappedObject).value = property.value
-        }
-
-        override fun reload(wrappedObject: M) {
-            property.value = accessor(wrappedObject).value
-        }
-
-        override fun resetToDefault() {
-            property.value = defaultValue
-        }
-
-        override fun updateDefault(wrappedObject: M) {
-            defaultValue = accessor(wrappedObject).value
-        }
-
-        override fun isDifferent(wrappedObject: M): Boolean {
-            val modelValue = accessor(wrappedObject).value
-            val wrapperValue = property.value
-
-            return modelValue != wrapperValue
-        }
-
-        override fun isValid(): Boolean {
-            return validator(property.value)
-        }
-
-        // TODO: do we also want to compare `property.name` or `defaultValue`?
-        @Suppress("UNCHECKED_CAST")
-        override fun equals(other: Any?): Boolean {
-            if (this === other) return true
-            if (other?.javaClass != javaClass) return false
-
-            other as FxPropertyField<*, *>
-            return accessor == other.accessor
-        }
-
-        override fun hashCode(): Int {
-            return accessor.hashCode()
-        }
-    }
-
-
-    /**
-     * An implementation of [PropertyField] that is used when the fields of the model class are **not** JavaFX
-     * Properties but are normal Kotlin properties (any visible `val` or `var`).
-     *
-     * @property accessor [KProperty1] or [KMutableProperty1] of the model class field
-     */
-    private inner class KPropertyField<T, out R : Property<T>>
-    //@JvmOverloads // causes CompilationException!
-    constructor(private val accessor: KProperty1<M, T>,
-            //private val setter: ((M, T) -> Unit)?,
-                private var defaultValue: T? = null,
-                propertySupplier: (M?, String?) -> R,
-                private val validator: (T) -> Boolean = { t: T -> true }
-               ) : PropertyField<T, M, R> {
-
-        override val property: R
-
-        init {
-            this.property = propertySupplier(null, accessor.name)
-            this.property.addListener { observable, oldValue, newValue -> this@ModelWrapper.propertyWasChanged() }
-        }
-
-        override fun commit(wrappedObject: M) {
-            if (isDifferent(wrappedObject) && accessor is KMutableProperty1<M, T>) {
-                accessor.set(wrappedObject, property.value)
-            }
-            // do nothing if property is immutable (`val`)
-        }
-
-        override fun reload(wrappedObject: M) {
-            property.value = accessor(wrappedObject)
-        }
-
-        override fun resetToDefault() {
-            property.value = defaultValue
-        }
-
-        override fun updateDefault(wrappedObject: M) {
-            defaultValue = accessor(wrappedObject)
-        }
-
-        override fun isDifferent(wrappedObject: M): Boolean {
-            val modelValue = accessor(wrappedObject)
-            val wrapperValue = property.value
-
-            return modelValue != wrapperValue
-        }
-
-        override fun isValid(): Boolean {
-            return validator(property.value)
-        }
-
-        @Suppress("UNCHECKED_CAST")
-        override fun equals(other: Any?): Boolean {
-            if (this === other) return true
-            if (other?.javaClass != javaClass) return false
-
-            other as KPropertyField<*, *>
-
-            if (accessor != other.accessor) return false
-            if (defaultValue != other.defaultValue) return false
-
-            return true
-        }
-
-        override fun hashCode(): Int {
-            var result = accessor.hashCode()
-            result = 31 * result + (defaultValue?.hashCode() ?: 0)
-            return result
-        }
-    }
-
-    /**
-     * An implementation of [PropertyField] that is used when the field of the model class is a [ListProperty] too.
-     *
-     * @param E
-     * *            the type of the list elements.
-     */
-    private inner class FxListPropertyField<E, T : ObservableList<E>, out R : Property<T>>
-    //@JvmOverloads // causes CompilationException!
-    constructor(private val accessor: KProperty1<M, ListProperty<E>>,
+                private val getter: (M) -> List<E>,
+                private val setter: ((M, List<E>) -> Unit)?,
                 private var defaultValue: List<E>? = mutableListOf<E>(),
                 private val validator: (T) -> Boolean = { t: T -> true }
-               ) : PropertyField<T, M, R> {
+               ) : PropertyField<T, M, R>() {
 
-        private val targetProperty = SimpleListProperty(FXCollections.observableArrayList<E>())
+        constructor(accessor: KProperty1<M, List<E>>,
+                    defaultValue: List<E>? = mutableListOf<E>(),
+                    validator: (T) -> Boolean = { t: T -> true }
+                   ) : this(accessor.name, accessor.getter,
+                if (accessor is KMutableProperty1<M, List<E>>) accessor.setter else null, defaultValue, validator)
 
-        init {
-            this.targetProperty.addListener(ListChangeListener<E> { change -> this@ModelWrapper.propertyWasChanged() })
-        }
+        constructor(name: String,
+                    accessor: (M) -> ListProperty<E>,
+                    defaultValue: List<E>? = mutableListOf<E>(),
+                    validator: (T) -> Boolean = { t: T -> true }
+                   ) : this(name, { accessor(it).value },
+                { m: M, l: List<E> -> accessor(m).value = FXCollections.observableArrayList(l) }, defaultValue,
+                validator)
 
-        override fun commit(wrappedObject: M) {
-            accessor(wrappedObject).setAll(targetProperty.value)
-        }
+        /*
+        constructor(accessor: KProperty1<M, ListProperty<E>>,
+                    defaultValue: List<E>? = mutableListOf<E>(),
+                    validator: (T) -> Boolean = { t: T -> true }
+                   ) : this(accessor.name, { accessor.get(it).value },
+                if (accessor is KMutableProperty1<M, ListProperty<E>>) { m: M, l: List<E> -> accessor(m).value = l}
+                else null as ((M, List<E>) -> Unit)?,
+                defaultValue, validator)
+         */
 
-        override fun reload(wrappedObject: M) {
-            targetProperty.setAll(accessor(wrappedObject).value)
-        }
-
-        override fun resetToDefault() {
-            targetProperty.setAll(defaultValue)
-        }
-
-        override fun updateDefault(wrappedObject: M) {
-            defaultValue = ArrayList(accessor(wrappedObject).value)
-        }
-
-        @Suppress("UNCHECKED_CAST")
-        override val property: R
-            get() = targetProperty as R
-
-        override fun isDifferent(wrappedObject: M): Boolean {
-            val modelValue = accessor(wrappedObject).value
-            val wrapperValue = targetProperty
-
-            return modelValue != wrapperValue
-        }
-
-        override fun isValid(): Boolean {
-            return validator(property.value)
-        }
-
-        @Suppress("UNCHECKED_CAST")
-        override fun equals(other: Any?): Boolean {
-            if (this === other) return true
-            if (other?.javaClass != javaClass) return false
-
-            other as FxListPropertyField<*, *, *>
-
-            if (accessor != other.accessor) return false
-            if (defaultValue != other.defaultValue) return false
-
-            return true
-        }
-
-        override fun hashCode(): Int {
-            var result = accessor.hashCode()
-            result = 31 * result + (defaultValue?.hashCode() ?: 0)
-            return result
-        }
-    }
-
-    /**
-     * An implementation of [PropertyField] that is used when the field of the model class is a [List] and
-     * is **not** a JavaFX [ListProperty].
-
-     * @param E
-     * *            the type of the list elements.
-     */
-    private inner class KListPropertyField<E, T : ObservableList<E>, out R : Property<T>>
-    //@JvmOverloads // causes CompilationException!
-    constructor(private val accessor: KProperty1<M, List<E>>,
-                private var defaultValue: List<E>? = mutableListOf<E>(),
-                private val validator: (T) -> Boolean = { t: T -> true }
-               ) : PropertyField<T, M, R> {
-
-        private val targetProperty = SimpleListProperty(FXCollections.observableArrayList<E>())
+        private val targetProperty = SimpleListProperty(null, name, FXCollections.observableArrayList<E>())
 
         init {
             this.targetProperty.addListener(ListChangeListener<E> { change -> this@ModelWrapper.propertyWasChanged() })
@@ -557,7 +370,7 @@ class ModelWrapper<M : Any>
         // need to have proper `equals()`-implementations. I'm not sure if we can assume this to be the case :/
         // Also, we may not change the order of the list... With a [Set] it's probably easier :/
         override fun commit(wrappedObject: M) {
-            val list = accessor(wrappedObject)
+            val list = getter(wrappedObject)
             if (list is MutableList<E>) {
                 list.clear()
                 list.addAll(targetProperty.value)
@@ -566,7 +379,7 @@ class ModelWrapper<M : Any>
         }
 
         override fun reload(wrappedObject: M) {
-            targetProperty.setAll(accessor(wrappedObject))
+            targetProperty.setAll(getter(wrappedObject))
         }
 
         override fun resetToDefault() {
@@ -574,7 +387,7 @@ class ModelWrapper<M : Any>
         }
 
         override fun updateDefault(wrappedObject: M) {
-            defaultValue = ArrayList<E>(accessor(wrappedObject))
+            defaultValue = ArrayList<E>(getter(wrappedObject))
         }
 
         @Suppress("UNCHECKED_CAST")
@@ -582,7 +395,7 @@ class ModelWrapper<M : Any>
             get() = targetProperty as R
 
         override fun isDifferent(wrappedObject: M): Boolean {
-            val modelValue = accessor(wrappedObject)
+            val modelValue = getter(wrappedObject)
             val wrapperValue = targetProperty
 
             return modelValue != wrapperValue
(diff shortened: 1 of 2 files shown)


edvin:
Nice! Give me a day or two to get back on my feet and I'll dig into it!

edvin:
Sorry for getting back to you so late on this. I've read most of the source code, but I don't feel that I have a complete grasp of the exact scope of the model wrapper, and some parts of it feels more involved than I feel it needs to be.

I wonder if if would be better to define exactly what this class should do, and then do a clean implementation to avoid unnecessary baggage and to make sure we do this as cleanly as possible by leveraging Kotlin to the fullest. I think we can do better syntax wise as well :)

I have time to experiment and write the implementation as well if needed. Can you help me with a definition of what you expect this class to do, then I can add some suggestions and we can start playing with an implementation?

edvin:
Actually, maybe I should create a small doc describing what I think the ModelWrapper is doing now and you can correct me? I don't want to waste your time and I feel that this will be very important for TornadoFX. I'm very glad that you brought this here :) I'll get back to you tonight (GMT+1).

johnp:
That would be great! I too feel like doing a more clean-room design will lead to something that can make more use of Kotlin and integrates better into TornadoFX. It was easier for me as a Kotlin-newbie and without much experience in library/framework-design to refactor the Java implementation for my use-case, so I'm happy if someone more experienced takes the lead ;)

edvin:
Here are some thoughts :)

Validation requires some more experimentation to get right, but can absolutely be added later, so I think we should skip it for now.

Minimal set of functions:

- Extract values from source object/objects
- Track `isDirty`
- Track `isDifferent`
- Commit values back to source object/objects
- Reset to initial values from source object/objects

The more I looked at your current implementation the more reasonable it seemed :) I don't like all the code for dealing with properties of different kinds, but there might not be a way around it.

In your SingleViewModel example you access the wrapper to create the field:

```kotlin
 val name = wrapper.field(Person::name)
```

I think we could make this:

```kotlin
val name = field(Person::name)
```

Looks nicer.

The reflection based view model is nice, but as long as we don't play around with byte code there isn't much to gain since you have to enumerate the fields anyway. Better to keep things explicit.

I think we can possibly just change the existing code a little, since most of it is pretty well written IMO :)

What pain points or issues you're seeing with the current implementation?

edvin:
The situation for TableView editing is a bit different, but I've been wanting to implement something like that as well. I remember ExtJS actually has a very nice inline table editor with dirty checking and commit etc. We need a separate issue for that however, because TableView actually provides almost everything we need for that already :)

edvin:
I have an idea that might reduce the complexity of the wrapper. I haven't thought this completely through, but I'll write it down so we can discuss it at least.

What if the wrapper only dealt with JavaFX Properties? Then it doesn't matter what domain object the properties are coming from, so we might not need different ViewModel bases classes depending on wether you have one or multiple domain objects as source.

We obviously have to support wrapping of non-JavaFX properties as well, but TornadoFX already has functions to turn a Plain Old Java Property into a JavaFX property, so that's already taken care if.

I think this means less code duplication and even easier handling of situations where the ViewModel handles properties from multiple sources.

Again, haven't thought this through, but if it's at all possible we have greatly reduced the complexity. Any thoughts?

edvin:
I created a very naive implementation of ViewModel just to illustrate what I'm talking about. Consider this:

```kotlin
open class ViewModel {
    val properties: MutableMap<Property<*>, Property<*>> = mutableMapOf()

    inline fun <reified T> Property<T>.wrap(): Property<T> {
        val wrapper = SimpleObjectProperty<T>(this.value)
        properties[wrapper] = this
        return wrapper
    }

    fun commit() {
        properties.forEach { it.value.value = it.key.value }
    }

    fun rollback() {
        properties.forEach { it.key.value = it.value.value }
    }

}
```

Now let's say we have a `Person` class with a name property and a `PersonViewModel`. I put `wrap` at the end here, it kind of seemed cleaner, don't know what you think?

```kotlin
class Person(name: String) {
    var name: String? by property(name)
    fun nameProperty() = getProperty(Person::name)
}

class PersonViewModel(person: Person) : ViewModel() {
    val name = person.nameProperty().wrap()
}
```

Now we can test the commit feature like this:

```kotlin
open class ViewModelTests {

    @Test fun simple_commit() {
        val person = Person("John")
        val viewModel = PersonViewModel(person)

        viewModel.name.value = "Jay"
        // person.name should still be "John"
        Assert.assertEquals(person.name, "John")

        // Commit and check that person.name was updated
        viewModel.commit()
        Assert.assertEquals(person.name, "Jay")
    }

}
```

edvin:
I added functionality so you can handle POJO properties like this:

```kotlin
val name = person.wrap(Person::getName, Person::setName)
```

johnp:
Regarding my pain points with the current implementation these were mostly based on the fact that I wasn't sure if this couldn't be written in a more Kotlin-native style and stylistic (functional/Kotlin-native vs OO) differences to the other parts of TornadoFX. Also my experimentation resulted in some overly complex and feature-ridden code (I have a version with per-property dirty/different-tracking for example and the InsaneViewModel really is insane and non-practical, but it taught me some nice reflection-tricks at least). It's good to get some comments so I can simplify things now :)


Just read your latest comment and some of it looks like a really good idea. Unfortunately the ``Person`` class wakes up some bad memories when I tried to use those property delegates (?) in my domain models and (JPA) annotations had to be applied to the getter, because they are not possible on delegates, which blew up the whole class to about thrice the size.  
Using ``inline fun <reified T> Property<T>.wrap(): Property<T>`` and specialized versions such as ``(internal) inline fun String.wrap(): StringProperty`` in the ViewModel in place of the ``field`` methods in the ModelWrapper is a nice idea, but only as long as the model doesn't change, whereas I see changing the underlying model instance as a crucial feature. Typical CRUD-case with a ``TableView`` and an according ``DetailsPane`` for example.  
Changing this to ``internal inline fun <reified T> K(Mutable)Property1<*, T>.wrap(): Property<T>`` would allow ``val name = Person::name.wrap()`` and also allow changing the underlying model instance.

The problem then becomes, that the base class doesn't have a field of the model and couldn't actually provide any functionality (it has a map of the accessors, but doesn't have an object to apply them to).
We could try to solve this by allowing the user to register a model-property, but this would also mean that the user cannot use two objects of the same model concurrently, because we wouldn't know upon which model the accessors should be called, although I think that this can be something that we may be justified to disallow then (we could "allow" this by allowing the user to manually provide an identifier, but I feel that would increase complexity needlessly, see below).

```Kotlin
open class ViewModel {
    val properties: MutableMap<KProperty1<*, T>, Property<*>> = mutableMapOf()
    val models: MutableMap<String, ObjectProperty<*>> = mutableMapOf() // only need String identifier to prevent object collision

    internal inline fun <reified T> KProperty1<*, T>.wrap(): Property<T> {
        val wrapper = SimpleObjectProperty<T>(this.value)
        properties[wrapper] = this
        return wrapper
    }

   internal inline fun <reified T: ObjectProperty<M>> fun register(modelProperty: T, identifier: String? = null) {
      models[identifier ?: modelProperty.value.javaClass.name] = modelProperty // sth like that
    }

   // without support for multiple instances of the same model
   //internal inline fun <reified T: ObjectProperty<M>> fun register(modelProperty: T) {
   //     models.add(modelProperty)
   // }

    fun commit() {
        properties.forEach {
          // search ``models`` for object of this class and call the setter on it
          // would have if this works without also remembering the KClass in ``properties``, because of type erasure
       }
    }
}
```
Not sure if register couldn't be somehow use kotlin delegates, to allow something akin to this:

```Kotlin
class PersonViewModel(
        val person: ObjectProperty<Person> by register()
    ) : ViewModel() {
    val name = Person::name.wrap()
}
```

It's quite late so I'll return with some fresh ideas tomorrow.

edvin:
I've made some changes so that the ViewModel can now wrap the following types of properties:

- JavaFX properties
- POJO getter/setter pairs
- Kotlin `var` properties

I've committed the ViewModel so you can take a look at it, but of course nothing is decided when it comes to API and functionality, just wanted to make it easier for you to check it out.

Look at [ViewModelTest](https://github.com/edvin/tornadofx/blob/master/src/test/kotlin/tornadofx/ViewModelTests.kt) to see how to wrap the 3 types of properties now:

```kotlin
// JavaFX Property
val name = person.nameProperty().wrap()

// POJO
val name = person.wrap(JavaPerson::getName, JavaPerson::setName)

// Kotlin var
val name = person.wrap(Person::name)
```

They all know about the owning object, so your concerns above should be unwarranted.

I don't see any issues vs JPA with this, do you? If so, do you have an example for me?

edvin:
I added dirty state tracking as well, so now this test passes:

```kotlin
    @Test fun var_commit_check_dirty_state() {
        val person = Person("John")
        val viewModel = PersonViewModel(person)

        Assert.assertFalse(viewModel.isDirty())

        viewModel.name.value = "Jay"
        Assert.assertEquals(person.name, "John")
        Assert.assertTrue(viewModel.isDirty())

        viewModel.commit()
        Assert.assertEquals(person.name, "Jay")
        Assert.assertFalse(viewModel.isDirty())
    }
```

johnp:
As far as I remember, annotating a class with JPA when using ``by property`` and ``getProperty`` turned out to be something like this

```Kotlin
class Person(id: Int, name: String) {
   private val idProperty by lazy { SimpleIntegerProperty() }
   fun idProperty() = idProperty
   var id: Int
      @Id
      get() = idProperty.get()
      set(value) = idProperty.set(value)

    private val nameProperty by lazy { SimpleStringProperty() }
     fun nameProperty() = nameProperty
     var name: String
        @Column
        get() = nameProperty.get()
        set(value) = nameProperty.set(value)
}
```

instead of something as simple as that (which is also why I find ViewModels so useful in that they separate the JavaFX properties from the domain model in a  very clean way)

```Kotlin
class Person(
    @Id
    var id: Int = 0,
    @Column
    var name: String = "") {
}
```

How would you go about changing the ``ViewModel`` model instance if e.g. a selection in a ``TableView`` changes? Like binding a tableviews ``selectionModel.selectedProperty()`` to the ViewModel?

edvin:
Yeah, I agree that in the case of JPA it might be better to use the clean approach in your domain object. Fortunately the ViewModel now supports these properties no problem, see the examples above.

I'm not sure why you would put the `selectionModel.selectedProperty()` in a ViewModel, since it's a `ReadOnlyBooleanProperty` that cannot be changed, can you explain the use case for that? In any case, you can just add it without wrapping, it makes no sense to track dirty state of something that is read only.

johnp:
How would you change the ``person`` in your current ViewModel?

(my ``TableView`` example means a ``TableView<Person>``, maybe that is better understandable)

edvin:
Ah, I see :) So you want to use the same ViewModel and swap the underlying object. Let me see...

edvin:
I have a theory, but it's going to take a little time to implement a test for it. I will try to get it done tonight and get back to you, OK?

johnp:
Yes :) Sorry for my weird explanation.

edvin:
Nah, it was all good, I just didn't think of this (very important) usage pattern :) I want to stress that I don't know if my proposal here is viable or the best way forward, but I'd like to investigate it until we either hit a dead end or find out that it's viable.

edvin:
I think I have it, with a slight syntax change.

```kotlin
class PersonViewModel(var person: Person) : ViewModel() {
    val name = wrap { person.nameProperty() }
}

@Test fun swap_source_object() {
    val person1 = Person("Person 1")
    val person2 = Person("Person 2")

    val viewModel = PersonViewModel(person1)
    Assert.assertEquals(viewModel.name.value, "Person 1")

    viewModel.person = person2
    viewModel.rebind()
    Assert.assertEquals(viewModel.name.value, "Person 2")
}
```

Is that what we need? I haven't commited yet, because I need to adapt the `var` and `getter/setter` approach first. Gotta run, will be back in a couple of hours.

edvin:
All tests are passing again, different property types are bound like this:

```kotlin
// JavaFX Property
class PersonViewModel(var person: Person) : ViewModel() {
    val name = bind { person.nameProperty() }
}

// Java POJO getter/setter property
class JavaPersonViewModel(person: JavaPerson) : ViewModel() {
    val name = bind { person.observable(JavaPerson::getName, JavaPerson::setName) }
}

// Kotlin var property
class PersonVarViewModel(person: Person) : ViewModel() {
    val name = bind { person.observable(Person::name) }
}
```

I'll see if I can tighten the syntax further. Do you see any problems with this approach? After changing the `person` instance you have to call `rebind()`. It would be nice to avoid this, but I don't think that's possible unless the var is defined as a special var somehow. Will investigate.

edvin:
Hm.. maybe ViewModel can accept a default model object and a way to change it.. that way we could reduce the syntax for the common case (SingleViewModel) and keep support for MultiViewModel in the same class. I'll investigate this as well.

edvin:
I slightly better way to do rebind:

```kotlin
    @Test fun swap_source_object() {
        val person1 = Person("Person 1")
        val person2 = Person("Person 2")

        val viewModel = PersonViewModel(person1)
        Assert.assertEquals(viewModel.name.value, "Person 1")

        viewModel.rebind {
            person = person2
        }

        Assert.assertEquals(viewModel.name.value, "Person 2")
    }
```

Still not perfect...

edvin:
I found a way to expose both a var (getter/setter) as well as a property in the ViewModel. That's super nice if you refer to the values from your ViewModel a lot in code, so you can write `viewModel.name = "John"` instead of `viewModel.name.value = "John"`. Using it is totally optional of course. It would look like this:

```kotlin
class PersonViewModel(var person: Person) : ViewModel() {
    var name : String by property { person.nameProperty() }
    fun nameProperty() = getProperty(PersonViewModel::name)
}
```

johnp:
Wow, that really looks quite neat! I first didn't recognize how some of it works but now that I do, after having spent some time on it, I'm quite impressed 👍 
The only thing I don't like is that we also use an ``ObjectProperty`` on the side of the model, but I think that can only be improved once we have bound property references, so then we could probably do this:

```
class PersonVarViewModel(person: Person) : ViewModel() {
    val name = bind { person::name } // lambda type:  () -> KMutableProperty0<String>
}
```

Note that that should also work with java classes, just like ``PersonVarViewModel`` should also work for ``JavaPerson``: https://kotlinlang.org/docs/reference/java-interop.html#java-reflection

Edit: Just tested it with Kotlin 1.1 and it works.

edvin:
Oh yes, property references will clean this up even further. Cool that you tested it with Kotlin 1.1, then we know that the future is bright for this approach :)

What do you think, should we try to go down this road? What more features do we need?

johnp:
I think we can and I also have a couple of things on my mind:

- Currently we manage 2 HashMaps and one ArraryList. We could put getter, setter, the ``propertyProvider`` (``supplier`` in ``ModelWrapper``) and a boolean ``dirtyFlag`` into a class like the ModelWrappers ``PropertyField``, so we'd be able to stay with just one ``HashMap<Property<*>, PropertyField<*>>`` (``PropertyField/BoundField/BackingField`` maybe). Not sure if  it's worth it though. We could then do this instead of the ``.observable()``, but it may be easier to just wait for property references:

```Kotlin
class PersonViewModel(person: Person) : ViewModel() {
    val name = bind({ person }, Person::name)
}
```

- IMHO we should overload ``bind`` with ``IntegerProperty`` and it's counterparts, as well as ``ListProperty`` and ``SetProperty`` to support ``IntegerExpression``, ``NumberBaseExpression``, ``ListExpression`` and so on, e.g.:

```Kotlin
    private fun <S : Property<T>, T> _bind(op: () -> S): Property<T> {
        val prop = op()
        val value = prop.value
        @Suppress("UNCHECKED_CAST")
        val wrapper = when(value) {
            is String -> SimpleStringProperty(value)
            is Int -> SimpleIntegerProperty(value)
            is Float -> SimpleFloatProperty(value)
            is Double -> SimpleDoubleProperty(value)
            is Boolean -> SimpleBooleanProperty(value)
            is ObservableList<*> -> SimpleListProperty(value)
            is ObservableSet<*> -> SimpleSetProperty(value)
            is List<*> -> SimpleListProperty(value.observable())
            is Set<*> -> SimpleSetProperty(value.observable())
            else -> SimpleObjectProperty<T>(value)
        } as Property<T>
        @Suppress("UNCHECKED_CAST")
        wrapper.addListener(dirtyListener as ChangeListener<in T>)
        properties[wrapper] = prop
        propertyProviders[wrapper] = op
        return wrapper
    }

    fun <S : Property<T>, T> bind(op: () -> S) = _bind(op)
    fun bind(op: () -> Property<String>) = _bind(op) as StringProperty
    fun bind(op: () -> Property<Int>) = _bind(op) as IntegerProperty
    fun bind(op: () -> Property<Float>) = _bind(op) as FloatProperty
    fun bind(op: () -> Property<Double>) = _bind(op) as DoubleProperty
    fun bind(op: () -> Property<Boolean>) = _bind(op) as BooleanProperty
    @Suppress("UNCHECKED_CAST")
    fun <E> bind(op: () -> Property<List<E>>) = _bind(op) as ListProperty<E>
    @Suppress("UNCHECKED_CAST")
    fun <E> bind(op: () -> Property<Set<E>>) = _bind(op) as SetProperty<E>
```

- Do we have a way to easily bind e.g. a ``(TableView.)selectionModel.selectedItemProperty()`` to the model?

```Kotlin
  (...).selectionModel.selectedItemProperty().addListener { sth, oldPerson, newPerson -> viewModel.rebind { person = newPerson ?: Person() }}
  // better ? what if ``selectedItemProperty() == null``?
  (...).selectionModel.selectedItemProperty().bind(viewModel.personProperty())
```

edvin:
Lots of good stuff here, great! The `bind` overloading in the proposed form is problematic because of type erasure. For example `Property<String>` and `Property<Double>` are erased to the same type, so if you have a `Property<String>` it would match almost all the overloaded functions. I'll look for a workaround.

edvin:
This is also problematic:

````kotlin
val wrapper = when(value) {
            is String -> SimpleStringProperty(value)
```

It would only return `SimpleStringProperty` if the property had a value at bind time. This would lead to confusing behavior.

edvin:
OK, so all the overloaded bind-functions are a no go because of type erasure, but I made sure the bind function returns the correct type so that if you bind a `Property<String>` you will get a `StringProperty` back etc. If you don't specify a type ourself, you get a `Property<String>`:

```kotlin
// name type will be Property<String> even though the instance is actually SimpleStringProperty
val name = bind { person.nameProperty() }

// name type is forced to be SimpleStringProperty
val name = bind { person.nameProperty() } as SimpleStringProperty
```

So basically, if you care about the type, for example because you need to interact with `IntegerExpression`, you would add the cast to `IntegerProperty` or `SimpleIntegerProperty`.

edvin:
I will find a way to reduce the number of objects created in the ViewModel and look at a sexy way to rebind when a property changes (for the selectionModel.selectedItem use case). It's getting late here and I want to get back up at 04:00 to catch UFC 200, so I'll play more with this tomorrow :)

edvin:
I managed to remove the `propertyProviders` map by making the value in the providers map a `Pair` of the Property and the PropertyProvider.

I don't know if it's a good idea to remove the `dirtyProperties` list. It's nice to subscribe to it to update your UI, and I'd have to keep the state of dirty properties somewhere, so I doubt an alternative approach would use less memory :)

edvin:
I added a suggestion for handling the TableView case:

The `rebindOnChange` function takes an observable it can listen to and extracts the value from it on each change. The value is sent to the `op` block supplied, which operates directly on the ViewModel. Simply assign the new value and `rebind()` will be called for you:

```kotlin
    @Test fun tableview_master_detail() {
        val tableview = TableView<Person>()
        tableview.items.addAll(Person("John", 37), Person("Jay", 33))
        val viewModel = MasterDetailViewModel(tableview.items.first())
        Assert.assertEquals(viewModel.name.value, "John")
        viewModel.rebindOnChange(tableview.selectionModel.selectedItemProperty()) {
            person = it ?: tableview.items.first()
        }
        tableview.selectionModel.select(1)
        Assert.assertEquals(viewModel.name.value, "Jay")
    }
````

Would that work OK?

Looking at `properties` I see that it has become a mess, so I'll implement your suggestion for a `PropertyField/BoundField/BackingField` or similar later today :)

**Comment by edvin** [comment]
Haha, I just realized that we don't need to store the actual property in the map, we just need the extractor function so we can get at it when we need to. Properties is now defined like this:

```kotlin
val properties = FXCollections.observableHashMap<Property<*>, () -> Property<*>>()
```

The `key` is the property facade and the value is the function that knows how to extract the property from the source object. Much simpler :)

edvin:
I also added a convenience function for `TableView` and `ListView` to support the selection case specifically:

```kotlin
viewModel.rebindOnChange(tableview) {
    person = it ?: Person()
}
```

edvin:
I have started writing some documentation that reflects my current understanding of the ViewModel:

https://github.com/edvin/tornadofx/wiki/ViewModel

edvin:
I added multiple convenience functions to check if a value is dirty and to extract the original value:

```kotlin
val sourceObjectValue = model.backingValue(model.name)
val isNameDirty = model.isDirty(model.name)
// Fancy version of the above so you don't need to repeat model all over the place
val isNameDirty = model.isDirty { name }
val isNamePristine = model.isNotDirty { name }
```

thomasnield:
I really need to play with this today it looks very useful for what I'm doing. I have some ideas but I want to play with this first. Is it possible to not make our own implementation of observable value that has dirty functions right on it?

edvin:
Great, let me know how it goes!

It's possible, but we would need to create a large number of subclasses because we need to support the extra functions exposed by `IntegerProperty`, `BooleanProperty`, `AllKindsOfNumberProperty` etc.

edvin:
@johnp What do you think so far? Thomas asked about validation and I think we could even squeeze that in for the first iteration. I'll start looking at it now.

johnp:
Happy with it so far. I haven't had a lot of time to look at it the last couple of days, but I'll probably be able to give some more feedback over the next days. One thing to note regarding validation is, that, in contrast to ``dirtyProperties``, one exposed ``ReadOnlyBooleanProperty`` per field would be quite helpful to indicate to the end-user which input field is invalid. It should also be considered, if validation only takes place during ``commit`` or with each change in input (expand ``dirtyListener``). The second one would be more desirable, but then it'd be nice to have a way of lazifying / delaying validation while the user e.g. still types text into a textfield. 

Example: Input ``textfield`` validation is triggered if value remains unchanged for 500ms or focus is lost. Then execute custom ``onValidation`` handler that for example opens a ``ControlsFX.PopOver`` with an error message. Possible syntax:

```Kotlin
textfield().bind(viewModel.name). validate({ /* validator */ }, { /* onValidation */ }, 500)
```

edvin:
Thanks for the feedback @johnp - Very valuable, I like the delay option. I will start thinking about this and try to come up with an implementation within the next few days.

edvin:
I see some validation frameworks allow a failed validation to be of a certain type, for example WARNING, ERROR or INFORMATION. I've never had use for that myself and could easily get by with a Boolean. What about you guys?

edvin:
About ControlsFX - they have already done stuff we would need to properly support validation and decoration. I will look at ways to support both integration with ControlsFX but also investigate how much work it would be to support something similar that we could include in TornadoFX.

johnp:
I just realized that validator and onValidation can actually be merged, so we wouldn't have to care about the return type. The above syntax wouldn't actually need the ``viewModel``, but would still need some way to be bound to be called before a ``commit`` can happen.

edvin:
Good call, I have an idea about how to do that.

edvin:
If we find that we only need a Boolean for validation, maybe the validator could simply return the error message or null if valid.. Or is that too simplistic? I like to start with the simplest approach possible and expand as needed :)

thomasnield:
I kind of like the ReadOnlyBoolean for each property idea,  as well as integrating with CotnrolsFX validation or at least doing something similar.

edvin:
I think we should be able to do decorations ourselves. It was quite easy to do this:



```kotlin
field("Username") {
    textfield {
        polygon(0.0, 0.0, 0.0, 10.0, 10.0, 0.0) {
            isManaged = false
            fill = Color.RED
        }
        tooltip("Invalid username, supply atleast 5 chars") {
            Platform.runLater {
                val input = this@textfield
                val b = input.localToScreen(input.boundsInLocal)
                show(input, b.minX, b.maxY)
            }
        }
    }
}
```

Shape builders for the win :)

johnp:
Nice! 
Regarding the delayed validation, I think we'd have to use [Timeline](http://docs.oracle.com/javase/8/javafx/api/javafx/animation/Timeline.html) and either cancel/restart the previous timeline (requires state keeping, which I'd like to avoid) or to validate if after X milliseconds the value is the same as before. This means e.g. a user could have typed a character and pressed backspace and then we'd still call the validator, although neither 500ms elapsed since the last input, nor the focus was lost, but it'd be possible without keeping track of any state short of the text in the field at the time the timeline started, which can be moved into the lambda during the creation of the timeline. I'm not sure if that's an acceptable shortcoming, but I can draft up a PoC and come back to you if you like.

edvin:
That would be helpful. I'll leave out the delay until you have had a chance to look at it :)

edvin:
Maybe it's better to trigger the validation on input focus lost?

johnp:
I'm not sure if I'm using ``Timeline.keyframe()`` right here and this is unfinished/untested, but it should be enough to get the idea:

```Kotlin
// todo: should validation be (TextField) -> Unit/Boolean ??
// validation == validator + onValidation
// we could change this to run validator in the background and onValidation on the UI thread if necessary, but then we'd have to split them up
inline fun TextField.validate(delay: Duration = Duration.ZERO, crossinline validation: (String) -> Unit): TextField {
    textProperty().addListener { observable, oldValue, previousValue ->
        Timeline().keyframe(delay, { // execute the following with ``delay`` delay on the UI thread
            val currentValue = this@validate.text
            if (currentValue != previousValue) {
                validation(currentValue)
            }
        })
    }
    focusedProperty().addListener { observable, previouslyFocused, currentlyFocused ->
        if (!currentlyFocused && previouslyFocused) {
            val currentValue = this@validate.text
            validation(currentValue)
        }
    }
    return this
}

inline fun TextField.validate(delay: Duration = Duration.ZERO, message: String, crossinline validator: (String) -> Boolean): TextField {
    validate(delay) {
        if (validator(this.text)) {
            // this.popOver(message, Styles::FIELD_VALIDATION_ERROR)
            if (this.tooltip != null) this.tooltip(message) // we don't want the tooltip to fade in multiple times, which would be weird
        } else this.tooltip = null // hide()
    }
    return this
}
```

gtnarg:
@edvin - it seems to me that the distinction between ERROR and WARNING could be important.  I have worked on applications where the form cannot be committed if ERROR(s) exist but WARNING(s) will not prevent it.

edvin:
@gtnarg That's a valid use case indeed. OK, we'll try to incorporate it.

johnp:
I made some fixes to my previous comment and ``validation`` should really be called with the textfield as argument, so users can easily implement warning/error distinction or more complicated UI/validation interaction themselves.

t-boom:
Why do you need a delay? From what I've seen, it's much more common to validate things like text boxes when they lose focus.

edvin:
Small proof of concept:

edvin:
Maybe we don't need the delay. Now you can do this:

```kotlin
validation.addValidator(this, trigger = ValidationTrigger.onblur) {
    if (it == "Edvin") error("That's a weird name") else
    if (it == "John") warning("Meeh...") else null
}
```

Now it will trigger when the textfield looses focus, or whenever validation.validate() is called. Also note that if you type "John", "Meeh" will be displayed as a warning, but `validation.validate()` or `validation.isValid` will still return true.

Both `warning` and `error` are functions that return the correct type of `ValidationMessage`:

```kotlin
fun error(message: String) = ValidationMessage(message, ValidationSeverity.Error)
fun warning(message: String) = ValidationMessage(message, ValidationSeverity.Warning)
```

thomasnield:
Wow this is pretty slick.

edvin:
Thanks :) I think it is both powerful and very easy to reason about for beginners. You can do some crazy stuff with it already, more examples coming tomorrow :)

johnp:
I agree. Could you try out 

```Kotlin
validation.addValidator(this, trigger = ValidationTrigger.onblur) {
    if (it.length < 7) error("Too short! ") else
    if (it.length < 10) warning("OKish") else null
}
```

OK, I think we are perfectly good without the delay. 😊

edvin:
That should work perfectly, @johnp :)

edvin:
I have some ideas about how to create a default context if one doesn't exist etc, so you can basically just do `addValidator` on any `Control`. Haven't decided if the model needs to be involved or not yet. Need to try some things tomorrow.

edvin:
The `when` operator is really nice for these kinds of things by the way:

```kotlin
when (it) {
    "Edvin" -> error("That's a weird name")
    "John"  -> warning("Meeh")
    else -> null
}
```

t-boom:
Quick question: Why is `ValidationTrigger` defined as
``` kotlin
sealed class ValidationTrigger {
    companion object {
        val onchange = OnChange()
        val onblur = OnBlur()
        val none = None()
    }
    class OnBlur : ValidationTrigger()
    class OnChange(val delay: Long = 0) : ValidationTrigger()
    class None : ValidationTrigger()
}
```
instead of
```
sealed class ValidationTrigger {
    object OnBlur : ValidationTrigger()
    object OnChange(val delay: Long = 0) : ValidationTrigger()
    object None : ValidationTrigger()
}
```
?

edvin:
Because constructors are not allowed for objects, so that code doesn't compile :)

I have OnChange delay working now:

```kotlin
validation.addValidator(this, OnChange(500)) {
    when (it) {
        "Edvin" -> error("That's a weird name")
        "John" -> warning("Meeh")
        else -> null
    }
}
```

johnp:
Nice! I don't see ViewModel integration right now, so one could commit the model even if there are validation failures. Pass ValidationContext to ViewModel or the other way round? Multiple validationContexts per ViewModel or not? Or should the ViewModel itself expose a validationContext instance?

t-boom:
@edvin Yeah, I meant to leave `OnChange` as a class. See my edit.

edvin:
I guess that's cleaner, even though we now create new OnChange elements all the time. Doesn't really matter I guess :) Maybe we should keep a default onchange instance with delay=0?

edvin:
Maybe:

```kotlin
sealed class ValidationTrigger {
    object OnBlur : ValidationTrigger()
    class OnChange(val delay: Long = 0) : ValidationTrigger()
    object None : ValidationTrigger()
    companion object {
        val OnChangeNoDelay = OnChange()
    }
}
```

So I can use the OnChangeNoDelay for all the default values?

edvin:
My sealed class foo is weak. Thanks for cleaning it up, @t-boom :)

t-boom:
Or even
``` kotlin
sealed class ValidationTrigger {
    class OnChange(val delay: Long = 0) : ValidationTrigger()
    object OnBlur : ValidationTrigger()
    object None : ValidationTrigger()
    object OnChangeNoDelay : OnChange()
}
```

t-boom:
I'm not sure if there would be an advantage to putting it on a companion object. (There might be something wrong with an `open` requirement in my idea. I need to sit down long enough to get IntelliJ up before I post ideas :) )

edvin:
I'm going to add an immediate boolean to each validator as well, so it can be evaluated at once.

edvin:
I think we might need an `AggregateValidator` as well, to be able to chain validators. Need to think about this.

edvin:
Added first steps of integration into ViewModel. Gotta run now :)

thomasnield:
Have a good night @edvin. This is looking awesome!

edvin:
I did a little bit on the integration part, so the following example is working. Time to clean up the syntax:

```kotlin
            val model = PersonModel(Person())

            fieldset("Customer Information") {
                field("Name") {
                    textfield(model.name) {
                        model.addValidator(this, model.name) {
                            when (it) {
                                "Edvin" -> error("That's a weird name")
                                "John" -> warning("Meeh")
                                else -> null
                            }
                        }
                    }
                    button("Save") {
                        setOnAction {
                            if (model.commit()) {
                                println("A OK")
                            } else {
                                println("Failed")
                            }
                        }
                    }
                }
            }
```

edvin:
If you in the above example remove the last parameter to `addValidator`:

```kotlin
model.addValidator(this)
```

You would now bind against the `TextField.textProperty()` instead of `model.name`. The end result would be exactly the same though. It's a subtle difference, and I wonder if it could get us into trouble down the line. Something to think about :)

edvin:
I managed to reduce a little boiler plate, so now it's:

```kotlin
                    textfield(model.name) {
                        model.name.addValidator(this) {
                            when (it) {
                                "Edvin" -> error("That's a weird name")
                                "John" -> warning("Meeh")
                                else -> null
                            }
                        }
                    }
```

The reason you can do `model.name.addValidator` is that I added an extension function on `Property<T>` that works for properties created by the `bind` delegate inside a ViewModel. The "magic" is simply that these properties have their `bean` property set to the `ViewModel`, so I can extract the model from there :)

johnp:
Just a note (based on your example): ```TextField.textProperty().observable`` holds the ``model.name`` instance too, because it was just bound there. Unfortunately, ``observable`` is private in ``StringPropertyBase``, so we'd probably have to use reflection if we want to use it (or some Kotlin trick I don't yet know).

edvin:
It is bound, but it's not the same instance. The bean you find via `TextField.textProperty().bean` will be the `TextField`, while the bean you find via `model.name.bean` will be the `ViewModel`. That's a crucial difference if you try to do `textfield.textProperty().addValidator()` - it will now fail with the following error message:

`The addValidator extension on Property can only be used on properties inside a ViewModel. Use validator.addValidator() instead.`

edvin:
Funny, I just figured it's possible to use the validator for dirty state tracking as well. This example combines a an input validator with a dirty state validator. Dirty but valid fields will render a blue polygon and no popup (because the message is empty):

```kotlin
                    textfield(model.name) {
                        model.name.addValidator(this) {
                            when (it) {
                                "Edvin" -> error("That's a weird name")
                                "John" -> warning("Meeh")
                                else -> {
                                    if (model.isDirty { name }) info("")
                                    else null
                                }
                            }
                        }
                    }
```

To make the dirty state go away on save you have to validate after commit though, so it feels a bit hacky:

```kotlin
if (model.commit()) {
    // save here, then clear dirty state by calling validate()
    model.validate()
}
```

johnp:
Could you try this out (in ViewModel):

```Kotlin
inline fun <reified T> addValidator(
            node: TextInputControl,
            trigger: ValidationTrigger = ValidationTrigger.OnChangeImmediate,
            noinline validator: ValidationContext.(T?) -> ValidationMessage?) {

        val field = node.textProperty().javaClass.getDeclaredField("observable")
        field.isAccessible = true
        val property = field.get(node) as Property<String>

        validationContext.addValidator(node, property, trigger, validator)
    }
```

When calling textField(model.name), the ``model.name`` instance is bound in ``TextField.bind`` -> ``Binding.bindStringProperty`` -> ``StringProperty.bind`` as follows:

```Java
public abstract class StringPropertyBase extends StringProperty {
     // ...
     private ObservableValue<? extends String> observable = null;

     // ...

     /**
     * {@inheritDoc}
     */
    @Override
    public void bind(ObservableValue<? extends String> newObservable) {
        if (newObservable == null) {
            throw new NullPointerException("Cannot bind to null");
        }
        if (!newObservable.equals(observable)) {
            unbind();
            observable = newObservable;
            if (listener == null) {
                listener = new Listener(this);
            }
            observable.addListener(listener);
            markInvalid();
        }
    }

    // ...
}
```

edvin:
Yeah, it seems that the observable is not filled in. If we make this, the syntax will be like this:

```kotlin
                    textfield(model.name) {
                        addValidator {
                            when (it) {
                                "Edvin" -> error("That's a weird name")
                                "John" -> warning("Meeh")
                                else -> {
                                    if (model.isDirty { name }) info("")
                                    else null
                                }
                            }
                        }
                    }
```

That's pretty sweet :) This compiles now, but since it can't extract the underlying observable it fails.

johnp:
Yeah, I know why: We actually call ``bindBidirectional`` and not ``bind`` on ``TextField.TextProperty``, which then causes ``TextField.helper`` to be set to a ``BidirectionalBinding`` from which we should be able to extract the property via ``getProperty1`` or ``getProperty2``.

edvin:
The helper field is inside TextFieldContent, not the TextField :(

edvin:
Nah, this is a wild goose chase :) I got to the helper, but it was of type SingleInvalidation. Let's regroup and try again tomorrow :)

johnp:
Already got something that at least doesn't crash:

```
    fun addValidator(
            node: TextInputControl,
            trigger: ValidationTrigger = ValidationTrigger.OnChangeImmediate,
            validator: ValidationContext.(String?) -> ValidationMessage?) {

        val stringProperty = node.textProperty()
        val field = stringProperty.javaClass.getDeclaredField("helper")
        field.isAccessible = true
        val helper = field.get(stringProperty) as ExpressionHelper<String>

        val field2 = helper.javaClass.getDeclaredField("changeListeners")
        field2.isAccessible = true
        val bindings = field2.get(helper) as Array<*>
        val binding = bindings[0] as BidirectionalBinding<String>
        
        val field3 = binding.javaClass.getDeclaredMethod("getProperty1")
        field3.isAccessible = true
        val property = field3.invoke(binding) as Property<String>

        validationContext.addValidator(node, property, trigger, validator)
    }
```

But yeah, it's time to get some sleep :)

edvin:
Wow, you got it :) I changed it a little, so now the syntax above actually works. We have to think about how robust this really is though.

```kotlin
@Suppress("UNCHECKED_CAST")
fun TextInputControl.addValidator(
        trigger: ValidationTrigger = ValidationTrigger.OnChangeImmediate,
        validator: ValidationContext.(String?) -> ValidationMessage?) {

    val stringProperty = textProperty()
    val field = stringProperty.javaClass.getDeclaredField("helper")
    field.isAccessible = true
    val helper = field.get(stringProperty) as ExpressionHelper<String>

    val field2 = helper.javaClass.getDeclaredField("changeListeners")
    field2.isAccessible = true
    val bindings = field2.get(helper) as Array<*>
    val binding = bindings[0] as BidirectionalBinding<String>

    val field3 = binding.javaClass.getDeclaredMethod("getProperty2")
    field3.isAccessible = true
    val prop = field3.invoke(binding) as Property<String>

    if (prop is Property<*> && prop.bean is ViewModel) {
        val model = prop.bean as ViewModel
        model.addValidator(this, textProperty(), trigger, validator)
    } else {
        throw IllegalArgumentException("The addValidator extension on TextInputControl can only be used on inputs that are already bound bidirectionally to a property in a Viewmodel. Use validator.addValidator() instead.")
    }
```

edvin:
I've added several enhancements, including a `required` shortcut, and I'm currently looking at ways to improve interop between View, ViewModel, Validation and Form. The thing is, it seems to work pretty seemlessly already :)

I'm experimenting with adding an overridable ViewModel field to UIComponent (View/Fragment):

```kotlin
open val model: ViewModel? = null
```

But I don't know if it will have any value.

edvin:
Most of what I've tried is working very good now, so we need to start looking for holes and use cases that we might have missed. I'll start incorporating this in my apps, if any of you have a chance to do the same (or just create fictive test cases) that would be great!

edvin:
I managed to add a shortcut so that all common controls can call `addValidator` without specifying the property of the control. This cleans up almost all use cases for Controls. Used with the builders, you simply do:

```kotlin
anycontrol(model.prop) {
    addValidator {
    }
}
```

Here is the code, add to it if you see any missing controls :)

```kotlin
@Suppress("UNCHECKED_CAST")
inline fun <reified T> Control.addValidator(trigger: ValidationTrigger = ValidationTrigger.OnChange(), noinline validator: ValidationContext.(T?) -> ValidationMessage?) {
    val property: Property<*> = when (this) {
        is TextInputControl -> textProperty()
        is ComboBox<*> -> valueProperty()
        is RadioButton -> selectedProperty()
        is CheckBox -> selectedProperty()
        is ChoiceBox<*> -> valueProperty()
        is ColorPicker -> valueProperty()
        is DatePicker -> valueProperty()
        is Labeled -> textProperty()
        else -> throw IllegalArgumentException("Don't know how to extract the value-property from control of type $javaClass. Use model.addValidator or report an issue to get your control supported.")
    }

    val model = property.getViewModel()

    if (model != null)
        model.addValidator(this, property as Property<T>, trigger, validator)
    else
        throw IllegalArgumentException("The addValidator extension on TextInputControl can only be used on inputs that are already bound bidirectionally to a property in a Viewmodel. Use validator.addValidator() instead or update the binding.")
}
```

edvin:
Hmm.. there are some issues with this, trying to iron them out now.

edvin:
Got it! Basically we have to add one function for each control type, but it's real easy:

```kotlin
inline fun <reified T> ComboBox<T>.addValidator(trigger: ValidationTrigger = ValidationTrigger.OnChange(), noinline validator: ValidationContext.(T?) -> ValidationMessage?)
        = addValidator(this, valueProperty(), trigger, validator)
```

I've added for all the controls I can think of.



Should `addValidator` simply be `validator`? Would that look even nicer?

johnp:
I'd vote for ``validator`` in this context (builder), because ``addValidator`` here implies multiple validators on the same control, for which I don't see a point (``when(it)`` is better and less confusing, but I could be missing sth). ``addValidator`` should IMHO only be used for classes like ``ValidationContext`` and ``ViewModel`` (things that actually "own" a list of validators).

Otoh, using it somewhere else would imply that calling ``validator`` overwrites a previously set validator...

edvin:
Sounds good. Actually, multiple validators are supported, but probably not a good idea if they use the same decorator, because they might overlay over each other. I'll update the syntax :)

edvin:
Done. That felt good:

```kotlin
field("Sex") {
    combobox(model.sex, Sex.values().asList()) {
        validator {
            if (it == Female) null
            else error("We don't serve your kind")
        }
    }
}
```

johnp:
Dumb question: ``ValidationTrigger.OnChangeOrBlur``?

edvin:
Does that make sense? If you have onChange, the validation would already have triggered before you blur. Or am I missing something?

johnp:
Tabbing through fields without changing them? If e.g. a ``model.name`` was bound before the validator has been added, and this value is invalid, no validation will occur unless someone modifies the field. I know it's a weird case and eagerly validating the field when the validator is added spams the user needlessly, unless we make an eager validation less obtrusive, for example by only blending in the polygon without the tooltip. 

Otoh the validation would happen on ``commit`` either way. So I take back my question.

edvin:
There might be use cases for this. Let's play with it and see what's needed :) Anything else we are missing @johnp ?

johnp:
Off the top of my head, nothing. Probably needs some tests, but I'll first finish converting my current app, to see if something else pops up unexpectedly. I'll rename this PR for future reference and delete my modelwrapper branch, but we can keep this PR open till everything settled down. It looks like we ended up with something far better than I could have ever imagined 😄

edvin:
Very good! I'll need a couple of days for further testing myself, and then I'll try to create a screencast before launch :)

edvin:
Thank you so much for this pull request @johnp - I think we ended up with something pretty strong here!

Issue #12, opened 2016-02-21
More elegant builders/extension functions for GridPane?


thomasnield:
Really liking the vision of this library so far. I'm kind of fond of using `GridPane` because it gives a lot of layout control. Although it doesn't add much boilderplate, I wonder if there is a better way than calling `apply()` on a `Node` and then a bunch of static `GridPane` manipulation methods (as shown in bold below). 

<pre>
class MainApp : App() {
    override val primaryView = MainView::class

    class MainView : View() {
        override val root = GridPane()

        init {
            with(root) {
                tabpane {
                    <b>apply {
                        GridPane.setVgrow(this,Priority.ALWAYS)
                        GridPane.setHgrow(this,Priority.ALWAYS)
                    }</b>
                    tab("Report", GridPane()) {
                        row {
                            <b>this += ReportTable.create().apply {
                                GridPane.setHgrow(this, Priority.ALWAYS)
                                GridPane.setVgrow(this,Priority.ALWAYS)
                            }</b>
                        }
                    }
                }
            }
        }
    }
}
</pre>

Perhaps we can add extension methods to `Node` that relate to the `GridPane`? But then again that might be confusing to see an extended `hGrow` property on a `Node` because it is not obvious it relates to the `GridPane`.

edvin:
I only added the `row` builder for GridPane, and quickly moved on to create a `Form` layout for most of the stuff I used GridPane for in the past. Take a look at (https://github.com/edvin/tornadofx-controls/#form-layout) for quick and dirty info. I'm working on writing documentation and a sampler app for the controls in the project (and polishing them), but the `Form` component is really shaping up :)

It will try to finish documentation and examples next week, and blog about this specific component, because I feel it solves many problems much better than `GridPane`. I'm going to add specific builder support for this Form component as well.

I actually thought about the same thing, adding extensions functions to GridPane, but they would seem kind of out of place, so I need to take a step back and look for a better solution for GridPane.

I'll let you know if I can think of a nicer way to solve this :)

thomasnield:
I'm curious to check it out, but I did come up with something. 

```
fun <T: Node> T.constraints(op: (GridPaneConstraint.() -> Unit)): T {
    val gpc = GridPaneConstraint()
    gpc.op()
    return gpc.applyToNode(this)
}

class GridPaneConstraint(
        var columnIndex: Int? = null,
        var rowIndex: Int? = null,
        var hGrow: Priority? = null,
        var vGrow: Priority? = null,
[...]

PR #125, opened 2016-06-16
Add vboxMargin, hboxMargin, vboxGrow, and hboxGrow helpers.

radicaled:
I'm actually not sure what the appropriate place for these helpers are -- I'm assuming its `Layouts.kt` since this does deal with layouts.

Somewhat mutilated example:

```kotlin
vbox {
    hboxGrow(Priority.ALWAYS)
    text("Header") {
        alignment = Pos.TOP_CENTER
        vboxMargin(bottom = 20.0)
    }
    listview(FXCollections.observableArrayList("Hello", "World"))
    buttonbar {
        vboxMargin(top = 10.0)
        button("Accept Entry")
        button("Reject Entry")
    }
}
```

### DIFF
--- a/src/main/java/tornadofx/Layouts.kt
+++ b/src/main/java/tornadofx/Layouts.kt
@@ -1,5 +1,6 @@
 package tornadofx
 
+import javafx.geometry.Insets
 import javafx.geometry.Orientation
 import javafx.scene.Node
 import javafx.scene.control.*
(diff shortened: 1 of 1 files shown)


thomasnield:
Thanks @radicaled.  I think we implemented this already via hboxCobstraints and vboxConstraints.

thomasnield:
See #12

thomasnield:
And this chapter in the guide. Sorry for short messages.  On my phone. 

https://edvin.gitbooks.io/tornadofx-guide/content/Part%202%20-%20Usage%20Basics/7.%20Builders%20Part%20IV%20-%20Layouts%20and%20Menus.html

radicaled:
Thanks, hadn't noticed. Not sure how I feel about the syntax, though.

```
hboxMargin(top = 5.0, bottom = 20.0)
```

feels a bit easier than

```
hboxConstraints {
  marginTop = 5.0
  marginBottom = 20.0
}
```

What about augmenting hboxConstraints?
```
hboxConstraints {
   setMargins(top = 5.0, bottom = 20.0)
}
```

Still a bit wordy, but less so. I guess I like the ability to define the margins all on one line, without having to go through the hassle of `Insets(0.0, 20.0, 0.0, 5.0)` and then trying to remember which positional argument corresponds to top/left/bottom/right.

(nice job on the book by the way, guys (and/or gals))

thomasnield:
Thanks! I'm hoping we'll wrap it up completely by the end of the summer. 

Btw noticed a mistake in the guide. I'll correct my error later. It should be this:

```kotlin
vbox {
    button("Button 1") { 
        vboxConstraints {
            marginRight = 20.0
        vGrow = Priority.ALWAYS
    }
    }
    button("Button 2")
}
```

Regarding the augmentation, I don't have too strong of a preference. @t-boom and @edvin what do you think?

t-boom:
I also don't really care much either way. I don't feel the current method is that bad.

edvin:
Sure it could be improved slightly but I don't think we should break backwards compatibility because of it. I also don't like multiple ways of expressing the same thing if we can avoid it.

edvin:
I also tend to use inline type safe styles instead of setting layout properties where applicable, so I don't really use these functions a lot :)

radicaled:
Fair enough!

Thanks for the feedback, guys!

PR #139, opened 2016-07-07
optimized singleAssign() utility method

vach:
Using single variable for the object state (hence no synchronization is required) volatile will guarantee both visibility and atomicity of state change...

Most probable use case is when we have lot of reads from field initialized trough singleAssign(), in that case old code does 2 volatile reads, new code will do only one read... + methods are smaller hence jit will easily chew this code etc...

### DIFF
--- a/src/main/java/tornadofx/Properties.kt
+++ b/src/main/java/tornadofx/Properties.kt
@@ -91,15 +91,7 @@ fun <S : Any, T : Any> observable(bean: S, propName: String, propType: KClass<T>
 inline fun <reified T : Any> Any.observable(propName: String) =
     observable(this, propName, T::class)
 
-enum class SingleAssignThreadSafetyMode {
-    SYNCHRONIZED,
-    NONE
-}
-
-fun <T> singleAssign(threadSafeyMode: SingleAssignThreadSafetyMode = SingleAssignThreadSafetyMode.SYNCHRONIZED): SingleAssign<T> =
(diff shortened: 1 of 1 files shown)


edvin:
Great, thank you :) @thomasnield Do you want to review this?

thomasnield:
Thanks @vach. One question... what if the initialized value is in fact null? I

thomasnield:
For TornadoFX purposes I think you are correct in that assigning a null value is unlikely. But then this doesn't support nullable types and it is technically misleading.  I don't know,  it is more concurrent though.  What do you think @edvin?

vach:
@thomasnield you cannot initialize it with null, becuase the set method signature defines not nullable T type, so kotlin compiler will give you compile error.

P.S. i kept behaviour consistent with previous code, method signatures for get/set were not changed...

vach:
Oh wait, let me think

vach:
@thomasnield yep you're correct, even if you define nullable type and set null, you'll get an exception on first attempt to read, so this is only for non nullable types. I cant think of a scenario where you want a value to accept valid null once and never be able to update it...

On the other hand this discourages you to use bad practice of nulls, i'd keep it for that. 
But if you think you might need to store valid null then discard this (maybe keep commented out).

thomasnield:
Nulls are not so much a bad practice in my opinion,  but rather something that needs to be managed safely,  which Kotlin does with nullable types. 

Technically, I think this should support nullable and non-nullable types just like every other generic class in Kotlin. Unless anyone else feels differently, I think we should keep as-is. But please don't hesitate if you find anything else that could be improved.

vach:
Sure, i agree didn't think about nulls :)

PR #143, opened 2016-07-13
Feature/property delegation

t-boom:


src/main/java/tornadofx/Properties.kt
```diff
@@ -170,4 +170,34 @@ private class UnsynchronizedSingleAssign<T> : SingleAssign<T> {
...
+operator fun FloatProperty.setValue(thisRef: Any?, property: KProperty<*>, value: Float) {
+    this.value = value
+}
+
+operator fun LongProperty.getValue(thisRef: Any?, property: KProperty<*>) = value!!
+operator fun LongProperty.setValue(thisRef: Any?, property: KProperty<*>, value: Long) {
+    this.value = value
+}
+
+operator fun IntegerProperty.getValue(thisRef: Any?, property: KProperty<*>) = value!!
```
edvin: Should these return `Integer?` instead of `Integer!`? What about null values?

t-boom: It returns `Int`. The 5 primitive properties (`DoubleProperty`, `FloatProperty`, `LongProperty`, `IntegerProperty`, and `BooleanProperty`) protect against nulls, which is why they needed special cases. If you try to set them to `null`, it is stored as 0 or `false`.

edvin: Ah :) I see! Perfect, are we ready to merge?

t-boom: I pushed up a change to clarify

t-boom: It should be ready

edvin:
Very nice! We should mention this alternative in the wiki, and maybe even use it as the default way of creating properties in the documentation. If we do however, we should consider adding the kotlin-reflect dependency by default. If you don't use this feature (or any other feature that requires it) you can simply remove it from your own project. Less surprises if it's included by default. What do you think?

t-boom:
For now it's probably enough to mention there is an easier to use alternative, but also mention that this ease of use comes with some drawbacks, including the `column(...)` issue you pointed out. If it proves nice enough down the road, we can look at making it the default method.

edvin:
Do you have a chance to update the Wiki/Guide? I'm so swamped with the ViewModel/Validation things, and I need to finish both the Gradle plugin for FXLauncher and a new version of the IDEA-plugin for TornadoFX :)

t-boom:
I can update the wiki. As far as the guide goes, I can put something in there, but you or @thomasnield will probably have to rewrite it. I'm not a writer by an stretch of the imagination.

edvin:
Haha, I'm not so sure, but OK :) Anything helps right now. I also have to finish the chapter I started on shapes in the guide..

t-boom:
I added a small blurb about it to [the wiki](https://github.com/edvin/tornadofx/wiki/Properties) and [the guide](https://github.com/edvin/tornadofx-guide/blob/master/Part%202%20-%20Usage%20Basics/5.%20Builders%20Part%20II%20-%20Data%20Controls.md). You can check them to make sure I explained it properly.

PR #144, opened 2016-07-14
Shortened some code

t-boom:
Not sure if it's easier to read/maintain this way


edvin:
Very nice. I went so many iterations testing things, so this wasn't written like that in one go. This is more in line with how the rest of the framework is written. Thank you.

Issue #148, opened 2016-07-26
Let App accept class parameters


t-boom:
When creating an app, the main app class is almost always:

```kotlin
class MyApp : App() {
    override val primaryView = MyView::class

    init {
        importStylesheet(MyStyles::class)
    }
}
```

Would it be a good idea (or even possible) to let `App` accept those values as parameters? Something like

```kotlin
open class App(primaryView: KClass<out View>? = null, vararg stylesheet = KClass<out Stylesheet>) : Application() {
    open val primaryView: KClass<outView> = primaryView ?: DeterminedByParameter::class

    init {
        stylesheet.forEach { importStylesheet(it) }
    }

    ...
}
```

Then you could use

```kotlin
class MyApp : App(MyView::class, MyStyles::class)
```

Of course, the old way would still work fine.

t-boom:
I'm not sure if this would actually provide any benefit. It just popped into my head.

edvin:
I like it a lot. We should embrace everything that reduces boiler plate!

t-boom:
Added

edvin:
Very nice! This release will be just about reducing boiler plate it seems :)

t-boom:
I didn't think about adding an icon. Should that be an option as well?

edvin:
That would be nice.

PR #150, opened 2016-07-26
Added ability to add icon to App in constructor

t-boom:
It would be possible to do this with a single constructor, but it would feel funny. If you put the icon after the stylesheets, you always have to say `icon = Image("path/to/image")` instead of just `Image("path/to/image")`, whereas if you put it before you have to do the same with the stylesheets. This way makes if feel a little cleaner.

### DIFF
--- a/src/main/java/tornadofx/App.kt
+++ b/src/main/java/tornadofx/App.kt
@@ -9,8 +9,8 @@ import kotlin.properties.ReadOnlyProperty
 import kotlin.reflect.KClass
 import kotlin.reflect.KProperty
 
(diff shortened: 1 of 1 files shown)


t-boom:
Unless someone would rather this be a single constructor (see the caveats above), this can be merged.

t-boom:
This is a supplement to #148

thomasnield:
Oops, sorry X O

thomasnield:
I was looking at this and accidentally pressed ENTER on the merge button somehow...

PR #151, opened 2016-07-27
set modelStage before showAndWait()

pg5:
set modelStage before calling showAndWait() so closeModal() can access the stage


edvin:
Thank you for this. I'm merging it and making some small modifications :)

PR #166, opened 2016-08-29
Add Region builder function

Finnerale:
Simply adds
```java
fun EventTarget.region(op: (Region.() -> Unit)? = null) = opcr(this, Region(), op)
```
to `Layouts.kt`


edvin:
Perfect, thank you :) I will add it to the CHANGELOG.md as well.

PR #168, opened 2016-08-31
Set modalStage before blocking on showAndWait

pg5:
This was fixed previously but the fix got deleted in commit 464fb1dd81b13c1ea3542aef38078604abb95a59


edvin:
Sorry about that, I must have inadvertently reverted this while I was hunting another issue with the modal dialog. Thanks!

PR #174, opened 2016-09-09
Make the methods 'bind' accepting nullable properties

slimaku:
Hello, 

Today, I've got some difficulties in a particular use-case when using TornadoFX. So, I've decided to propose you a little improvement.

Please consider this model class :

``` kotlin
class MyClass {
    val dateProperty = SimpleObjectProperty<Date?>(null)
    val date: Date? by dateProperty
}
```

Because the type of the property is `Date?` and not `Date`, it was impossible to use the binding as described in the wiki : https://github.com/edvin/tornadofx/wiki/Binding.

By this pull-request, I propose to make possible to write `label().bind(dateProperty)` even if the binded property has a nullable generic type.

Please note this involves a refactoring of the function `getDefaultConverter`, especially in the manner of dealing with java primitive types. I've written two unit tests to ensure there would be no trouble with it.

Thank you for your reading and for your future feedback :-)

### DIFF
--- a/src/main/java/tornadofx/Binding.kt
+++ b/src/main/java/tornadofx/Binding.kt
@@ -17,6 +17,8 @@ import java.time.LocalDateTime
 import java.time.LocalTime
 import java.util.*
 import java.util.concurrent.Callable
+import kotlin.reflect.KClass
+import kotlin.reflect.jvm.jvmName
 
 inline fun <reified T> ComboBoxBase<T>.bind(property: Property<T>, readonly: Boolean = false) =
     if (readonly) valueProperty().bind(property) else valueProperty().bindBidirectional(property)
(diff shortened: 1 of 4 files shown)


thomasnield:
That is sensible. I'll let @edvin review and merge if he's cool with the changes. Thanks for the fix.

t-boom:
`SimpleObjectProperty<Date?>` is unnecessary. Since JavaFX properties are Java classes, `SimpleObjectProperty<Date>` can hold a null value. The only ones that can't are the primitive properties.

slimaku:
Yes, `SimpleObjectProperty<Date>` can hold null values. But if written in Kotlin, even if it could work, there is a semantic difference.

`SimpleObjectProperty<Date?>` would mean that a null value is possible and intended by the developer. Where `SimpleObjectProperty<Date>` would mean that there is no null value acceptable. (Even if in fact it would be accepted)

But I agree as you've pointed out, that it is possible to deal with nullable properties even without these changes. Thank you.

It made me think, and I have a question : 
Would it be better to let this code unchanged and instead allow to create such binding to `PropertyDelegate` instead ? And therefore my model class could be writted like this :

```kotlin
class MyClass {
    val dateProperty = property<Date?>(null)
    val date: Date? by dateProperty
}

// in a view :
label().bind(dateProperty)
```

edvin:
Thanks for bringing this up Jonathan. I think @t-boom is the authority on properties, what do you think Matthew?

t-boom:
@slimaku I see where you are coming from. Can you test another use case for me? In the case that you are using `SimpleObjectProperty<Int?>' and set it to `null`, does the string converter still handle it correctly? And does it work for `null` in `SimpleObjectProperty<Int>'?

t-boom:
Also, does the following look too confusing to you (despite the fact that it does the same thing:

```kotlin
class MyClass {
    val dateProperty = property<Date>(null)
    val date: Date? by dateProperty
}

// in a view :
label().bind(dateProperty)
```

or even:

```kotlin
class MyClass {
    val dateProperty = SimpleObjectProperty<Date>(null)
    val date: Date? by dateProperty
}

// in a view :
label(dateProperty)
```

By that I mean: is the code readability improved enough to justify the extra reflection callback?

(I'm genuinely curious here. It doesn't seem any more readable to me, but I recognize my familiarity with the code makes me biased.)

Everyone else's input is welcome as well of course :)

t-boom:
We actually have briefly discussed adding some sort of `KotlinObjectProperty<Type>` that isn't nullable. It would work similar to JavaFX's primitive properties that would return a default value if it were ever set to null. The default used would be whatever is passed in to the constructor (it would not be optional like other JavaFX properties).

slimaku:
I've added the test you asked, @t-boom. In the case of a null value , the integer-string converter give an empty string.

For me both are equaly readable for the property declaration. And I would slightly prefer to bind with `label().bind(dateProperty)` than `label(dateProperty)`, but both are ok.

The only thing I would find more valuable is to create the property with : 
`property<Date?>(null)` or `SimpleObjectProperty<Date?>(null)` or `KotlinObjectProperty<Date?>(null)` because in kotlin, if I use the generic type `<Date>` in kotlin, it's like announcing it cannot be null. (Even if it can in practice)

Actually, I find that the idea of using a `KotlinObjectProperty<Type>` whould be the best. And I woul'd be happy to propose an implementation, if you want :-)

edvin:
I totally agree. Please take a stab at implementing the `KotlinObjectProperty`. I think the implementation should be straight forward.

I do also see the value of allowing `Date?` for the bind function. How you define your properties is a matter of developer preference, and we might as well cater for both styles, since there is not much of a penalty to pay for allowing this.

slimaku:
Ok, I'll code a `KotlinObjectProperty` class this week.

edvin:
Perfect :) We are currently wrapping up the rest of the features for 1.5.6, so we'll hold off on the release until we can get this in there as well. This is a nice addition to the framework that both adds value and removes some confusion surrounding nullable types. Thanks a lot for your contribution.

t-boom:
@slimaku I've been digging into this more and am really starting to see the value with this change.

I have notice one other place this change would be appropriate, and I was wondering if you would like to make the change (it should be pretty much the exact same thing) as part of this PR in `ViewModel.bind`. (Also make sure other functions in there such as `property` are also updated appropriately.) If not, that's perfectly understandable.

t-boom:
As far as the `KotlinObjectProperty` class goes, it may be as simple as

```kotlin
class KotlinObjectProperty<T: Any>(val defaultValue: T, bean: Any? = null, name: String = "") : SimpleObjectProperty<T>(bean, name, defaultValue) {
    override fun get() = super.get() ?: defaultValue
    override fun set(newValue: T) = super.set(newValue)
    override fun setValue(v: T) = super.setValue(v)
}
```

but I haven't tested that thoroughly to see if it would cover everything.

slimaku:
I've created an another pull request (#176) with an implementation of a `KotlinObjectProperty`.

I will look for the `ViewModel.bind`

t-boom:
@slimaku Did you have a chance to look at `ViewModel.bind`?

slimaku:
No, I'm sorry I did't have any free time the last two days. I'll try to take the time before the weekend.

edvin:
Np, @slimaku - take your time, we want to get this right :)

t-boom:
@slimaku Agreed. I'm not trying to rush you, just trying to keep tabs on where everything is. Thanks for your help.

slimaku:
@t-boom No problem, I've well understood that. Thank you.

slimaku:
Okay, I've made the same change in the `ViewModel.bind()`.

PR #176, opened 2016-09-11
Add a KotlinObjectProperty class

slimaku:
In the MR #174, we discussed, that it would be nice to have a `KotlinObjectProperty` that can handle null value. Here is my implementaiton proposition.

How to use it :
```kotlin
class MyClass {
    val dateProperty = KotlinObjectProperty<Date?>(null)
    val date by dateProperty    // the inferred type of date is Date?
}
```

It's almost like a `SimpleObjectProperty` except, the property type is declared in kotlin, that allow more explicit typing and let the kotlin compiler ensure the absence of [NPE](https://kotlinlang.org/docs/reference/null-safety.html).

For exemple, this code will not compile :
```kotlin
class MyClass {
    val numberProperty = KotlinObjectProperty<Int>(42)
    var number: Int? by numberProperty // The kotlin compiler will raise a type mismatch (Int? is not compatible with Int)
}
```

### DIFF
--- a/src/main/java/tornadofx/Properties.kt
+++ b/src/main/java/tornadofx/Properties.kt
@@ -25,6 +25,20 @@ class PropertyDelegate<T>(val fxProperty: Property<T>) : ReadWriteProperty<Any, 
 
 }
 
+class KotlinObjectProperty<T>(
+        initialValue: T,
+        bean: Any? = null,
+        name: String = ""
+): SimpleObjectProperty<T>(bean, name, initialValue), ReadWriteProperty<Any, T>{
+
+    override fun get(): T = super.get()
+    override fun set(newValue: T) = super.set(newValue)
+    override fun setValue(v: T) = super.setValue(v)
+
+    override fun getValue(thisRef: Any, property: KProperty<*>) = get()
+    override fun setValue(thisRef: Any, property: KProperty<*>, value: T) = set(value)
+}
+
 fun <T> Any.getProperty(prop: KMutableProperty1<*, T>): ObjectProperty<T> {
     // avoid kotlin-reflect dependency
     val field = javaClass.findFieldByName("${prop.name}\$delegate")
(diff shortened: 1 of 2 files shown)


slimaku:
@t-boom, You will notice, it's pretty similar of the implementation you proposed. Except I don't use a default value. In fact this is only about typing. The generic type can be nullable but that's all.

A class that could be more close of what you thought could now be : 
```kotlin
class DefaultObjectProperty<T : Any>(val defaultValue: T, bean: Any? = null, name: String = "") : KotlinObjectProperty<T?>(null, bean, name) {
    override fun get(): T = super.get() ?: defaultValue
}
```

Should I add it in the pull request ?

slimaku:
Please note, that it still impossible to bind such properties with something like : `label().bind(myProperty)`.

But the MR #174 could be a solution to that problem.

edvin:
Thank you for this :) Let's play with it a little and decide whether we need the `DefaultObjectProperty` or not.

t-boom:
Wait, the point of the `KotlinObjectProperty` is to be ***non***-nullable. `SimpleObjectProperty` is already nullable. That's why the default value was needed: If the property ever gets set to null it returns the default value instead. Like how when you set a `SimpleIntegerProperty` to null it returns 0, or set `SimpleBooleanProperty` to null and it returns false. The point for a `KotlinObjectProperty` is to bring Kotlin's null safety to JavaFX properties.

t-boom:
See https://github.com/edvin/tornadofx/pull/174#issuecomment-246162228

slimaku:
Excuse me @t-boom I misunderstood. For me the point was able to declare nullable properties with `KotlinObjectProperty<Int?>` Instead of `SimpleObjectProperty<Int>`.

Thank you for creating an issue to talk about this subject.

t-boom:
It seems we all had different ideas of what this should mean :)

Issue #184, opened 2016-09-17
TextField Binding of Integer Property


gtnarg:
`textfield( model.myInt )`

where **model.myInt** is an Integer property is not supported whereas the following is:

```
  textfield(){
        bind( model.myInt )
   }
```

Is this by design or an oversight?

See full sample below.

Thanks
Grant

```

import javafx.beans.property.SimpleIntegerProperty
import tornadofx.*


class MyView : View("My View") {

    val model = MyPojoModel( MyPojo(1) )


    override val root = stackpane {

        // not supported
        // textfield( model.myInt )

        // works
        textfield(){
            bind( model.myInt )
        }

    }

    class MyPojo( var myInt: Int )

    class MyPojoModel(val mypojo: MyPojo) : ViewModel() {
        val myInt = bind { SimpleIntegerProperty(mypojo.myInt) }
    }

}
```

edvin:
It is by design, you must add IntegerStringConverter() as the second parameter. We might be able to support it though, I will check when I get back home tomorrow and let you know :)

bekwam:
Hi,

For the two-argument function, use NumberStringConverter.  javafx.beans.property.IntegerProperty is actually IntegerProperty<Number> rather than IntegerProperty<Integer>

        textfield(SimpleIntegerProperty(101), NumberStringConverter())

edvin:
Yeah, sorry, that's what I meant :) I'm on my phone in Oslo, just winging it :)

edvin:
This feature is now merged with master. Thanks @bekwam :)

PR #185, opened 2016-09-17
[#184] Added Property<Number> textfield() function

bekwam:
I added a method alongside the Property<String> textfield() method to handle Property<Number> arguments.  This will allow

textfield(SimpleStringProperty("test"))

as well as

textfield(SimpleIntegerProperty(101))
textfield(SimpleDoubleProperty(6.02))

I also put in a unit test "Controls.kt".  It initializes the FX toolkit, so it may be a little slow for the project.

### DIFF
--- a/src/main/java/tornadofx/Controls.kt
+++ b/src/main/java/tornadofx/Controls.kt
@@ -1,7 +1,6 @@
 package tornadofx
 
 import javafx.beans.property.Property
-import javafx.beans.property.SimpleIntegerProperty
 import javafx.beans.value.ObservableValue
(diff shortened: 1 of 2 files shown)


edvin:
Great work! Thank you :) I will review and merge tomorrow when I get home unless someone beats me to it :)

thomasnield:
Glad to find this targets all `Number` types. I'll go ahead and merge.

thomasnield:
Thought about adding convenience for `BigDecimalStringConverter` since I use that somewhat often for financial types. Then again, you might as well just explicitly pass the converter using the native methods.

PR #200, opened 2016-10-27
Post helper supports InputStream

HuXu77:
HttpURLEngine already supports using the InputStream, added a high level post function that accepts InputStream as a parameter.


edvin:
Thank you for this :) I have merged it now.

PR #212, opened 2016-11-29
added textArea property -fx-text-box-border

polmabri:


t-boom:
This one actually raises an interesting questions. `-fx-text-box-border` is not part of the JavaFX CSS specification, but it is defined in Modena, so **almost** all applications will have it. If the user sets the base style sheet to something else (Caspian for example), this and other such properties may not be defined (not a big deal) or could be defined in a different way (potentially a show-stopper requiring custom properties or forced/raw output).

Should we include such properties and assume everyone will use a stylesheet based of Modena (covering probably >99% of use cases)?

edvin:
Yes, I think it's a good call to include it. Modena is the default, so not including it would mean that it wouldn't be possible to style all facets of Modena without this. Also, as we have seen, the CSS docs are far from complete, so we must assume that people doing advanced styling knows how to check what css classes are applied to a certain node, either by inspecting the source or using scenic view or the layout debugger in TornadoFX. I vote go for it :)

bekwam:
Will there be a default stylesheet change with Java 9 the way there was from FX2 (caspian) -> 8 (modena)?

t-boom:
@bekwam I haven't heard of one. You'd have to ask Jonathan Giles.

edvin:
I'm pretty sure there won't be a modena replacement in JDK9, at least not with the JDK9 release.

edvin:
@t-boom Do you agree about merging this?

t-boom:
Sounds good to me.

We will have to keep this in mind for the future though if a replacement does ever come out. (It'll probably just be a matter of letting the user define their own `PropertyHolder` and creating built in options for the different defaults).

edvin:
Absolutely :) Merged!

PR #227, opened 2017-01-06
Param default value support

gtnarg:
Added default value support for param delegate.

`val booleanParam : Boolean by param( defaultValue = true )`

The delegate will also throw an IllegalStateException indicating which required parameter is missing (instead of a NullPointerException).

Also added builder support for adding fragments.

Open to feedback :)


edvin:
Very nice! I'll run a couple of tests and merge it :)

PR #229, opened 2017-01-12
overload for di() delegate to support named dependencies

gtnarg:
I use spring for DI and would like to be able to reference a bean by name (in addition to class).  This change allows me to configure the DiContainer (from java) as:

```
FX.setDicontainer(new DIContainer() {

	@Override
	public <T> T getInstance( KClass<T> type ) {
		return (T)getContext().getBean(  ((KClassImpl<T>) type).getJClass() );
	}

	@Override
	public <T> T getInstance( KClass<T> type, String name ) {
		return (T)getContext().getBean( ((KClassImpl<T>) type).getJClass(), name );
	}

});
```

which allows me to do the following:

```
val myBean: MyBean by di("myBeanName")
```

I added the additional method so as not to be a breaking change (from Java).  As there may not be many people for whom this is an issue, you could also just overload the existing `getInstance` method as well if you think that's a better approach.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## [1.5.10-SNAPSHOT]
 
+- di delegate overload to support injecting a dependency by name (in addition to type)
 - Fixed scope support. `DefaultScope(MyController::class)` or `MyController::class.scope(DefaultScope)`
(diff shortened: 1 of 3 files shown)


edvin:
Great addition indeed. You've taken care not to break backwards compatibility with the interface by providing a default implementation for the new version, so this looks good to me.

PR #236, opened 2017-01-22
Simplified/Enhanced Pojo Binding support

gtnarg:
Can now bind to a pojo (that follows java beans conventions) by only specifying a getter.

`myPojo.observable(MyPojo::getName)`

Uses javafx.beans.property.adapter.JavaBeanObjectPropertyBuilder under the covers.  An added benefit to this is that it now listens to propertyChangeEvents on the underlying pojo.

Unfortunately it is an api breaking change in that it returns ObjectProperty<T> and not PojoProperty<T> as before.  The way binding is usually used, this shouldn't generally cause any issues.  I understand if you decide not to implement this for this reason.

In addition, I couldn't figure out any way to make the following method backward compatible so I left it as is.

```
/**
 * Convert an bean instance and a corresponding getter/setter reference into a writable observable.
 *
 * Example: val observableName = observable(myPojo, MyPojo::getName, MyPojo::setName)
 */
fun <S : Any, T> observable(bean: S, getter: KFunction<T>, setter: KFunction2<S, T, Unit>): PojoProperty<T> {
    val propName = getter.name.substring(3).let { it.first().toLowerCase() + it.substring(1) }

    return object : PojoProperty<T>(bean, propName) {
        override fun get() = getter.call(bean)
        override fun set(newValue: T) {
            setter.invoke(bean, newValue)
        }
    }
}
```

Looking forward to your feedback and I'm fine with it if you don't want to implement at this point.

Grant

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 ## [1.6.1-SNAPSHOT]
 
+- can now bind to a pojo by providing only a single getter ( eg. person.observable( JavaPerson::getId ) )
+  - API break: previously returned a PojoProperty - now returns an ObjectProperty<T>
+  - uses javafx.beans.property.adapter.JavaBeanObjectPropertyBuilder and will now propogate PropertyChangeEvents from the pojo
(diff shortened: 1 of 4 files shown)


edvin:
This is great work! I don't think the change to return ObjectProperty is a big deal. I don't see any issues with this, so I'll merge right away. Thank you Grant!

PR #238, opened 2017-01-25
Feature - Color.derive and Color.ladder

t-boom:
The ladder isn't super efficient as it creates a linear gradient to normalize the stops (`Stops.normalize()` is package private, so it was the easiest way to access it that will be Jigsaw complaint), but I'm not sure if it's a big deal since it's only run once.


edvin:
That's a piece of art, Matthew :) No, I don't think the use of the linear gradient matters at all.

PR #241, opened 2017-01-28
fixed the dokka-maven-plugin error, by including jcenter as a plugin …

ronsmits:
…repository


edvin:
Thank you Ron! I'll try to remove the "javadoc" trap and see if we can satisfy the maven central sync requirements without it now.

PR #248, opened 2017-02-11
Fix TableView sort when using SortedFilteredList

Eliote:
When using SortedFilteredList.bindTo() to bind it to a TableView the header becomes "unclickable" and therefor unsortable. Setting tableView.items to the underlying "sortedItems" instead of "this" fix the problem.


edvin:
Thank you for this :) We'll release 1.6.2 with this fix some time next week.

PR #250, opened 2017-02-14
BROKEN: added kotlin controllers to fx:include test case

bekwam:


edvin:
I'll close this since we reverted the nested fxml support for now.

Issue #251, opened 2017-02-17
Autocomplete extension for combobox


aymen083:
Hi,

I created an extension for combox to make it auto completable. It may be interesting to integrate it into tornadoFX. I already use it in a tornadofx project.

This is a link to the code of that extension coded in kotlin : 
https://gist.github.com/aymen083/9c605e1b22ed36e0e48c5f46ac925948

I am not sure that is the best way to do in kotlin may be there is more efficient way to do. I am not an expert in kotlin language.

Thanks
Aymen

edvin:
Hi, and thank you for this.

It looks great!. I'd like a callback where you can return items programmatically based on the current input, and maybe a default implementation that operates on a list like in your example. We would also need a builder for this, with overrides to support a custom callback or just a list. (I can add the builder if you want).

Please file a PR and I'd be happy to include it!

aymen083:
I add comments with modification patch to pull request.
I don't know if you want to close this issue and continue the discussion on the pull request.

edvin:
Just checking in :) I'm having some success with creating a custom skin. I'm using most of your code, but also extending `ComboBoxPopupControl`. I think this might work. Will report back soon.

aymen083:
Very good.

edvin:
I added support for custom `cellFormat` as well, but I don't think it makes sense to support `buttonCell`, since we rely on the TextField to be able to do autocomplete. Does that make sense? It might still be more issues we need to deal with though. Do you have a chance to check it out and see how it works for you now?

aymen083:
I made a look at the implementation. 

It looks ok, but i think we should make a test on the used skin to not alter combobox with modified skin. Maybe add a comment in the documentation that the autocomplete feature modifier the skin of combobox will be sufficient.

I think also that we should add the support for visibleRowCountProperty.

And i found an interesting approach that add filter to combobox without making it editable this can be an enhancement to the actual implementation. It uses a tooltip to show the actual filter text ;
https://github.com/jesuino/javafx-combox-autocomplete

I don't have time right now but i will see if i can make a pull request based on your code to add this feature if the combobox is not editable.

I also wanted to change the implementation to use a FiltredList instead of the filter function which create an array in every call. My idea was to update the predicate in the filteredList after every text modification so the default filter implementation maybe  like this : 
```kotlin
val filtredItems: FilteredList<T> = (comboBox.items ?: FXCollections.emptyObservableList()).filtered { true }
var localFilter = true
[...]

PR #254, opened 2017-02-19
Autocomplete extension for combobox

aymen083:
Pull request for the issue #251 
I added a callaback and i made tests in the ComboBoxTest.kt

For the builder i did get how to do it.

Aymen Nouira

### DIFF
--- a/src/main/java/tornadofx/AutoCompleteComboBoxExtension.kt
+++ b/src/main/java/tornadofx/AutoCompleteComboBoxExtension.kt
@@ -7,17 +7,31 @@ import javafx.scene.input.KeyEvent
 import javafx.util.Callback
 
 
-fun <T> ComboBox<T>.makeAutocompletable() : AutoCompleteComboBoxExtension<T> = AutoCompleteComboBoxExtension<T>(this)
+/**
+ * Extension function to combobox to add autocomplete capabilities
+ * Default filter use the string produced by the converter of combobox and search with contains ignore case the occurrence of typed text
+ */
+fun <T> ComboBox<T>.makeAutocompletable() = AutoCompleteComboBoxExtension<T>(this, null)
+/**
+ * Extension function to combobox to add autocomplete capabilities
+ * Accept in parameter a callback to create the autocomplete list based on input text
+ * Default filter use the string produced by the converter of combobox and search with contains ignore case the occurrence of typed text
+ */
+fun <T> ComboBox<T>.makeAutocompletable(autoCompleteFilter : Callback<String, ObservableList<T>>) = AutoCompleteComboBoxExtension<T>(this, autoCompleteFilter)
 
 
 
 /**
+ * Auto Complete support for combobox
+ * Accept a call back to make custom filter
+ * Default filter use the string produced by the converter of combobox and search with contains ignore case the occurrence of typed text
  * Created by anouira on 15/02/2017.
  */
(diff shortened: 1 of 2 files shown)


edvin:
Great, thank you. I'll merge and add the builder. I will also change the use of the Callback. This JavaFX construct just makes it awkward to use in Kotlin, so a function is a much better option to make the interface nicer.

edvin:
You create an observable arraylist which is then thrown away for each filter invocation. I'll change it so it just returns a list, filtered with the Kotlin collections `filter` function instead :)

There is also an issue:

- If you drop the list and it shows a single or even no entries, then you change the selection so it contains multiple entries -> the list area doesn't expand. I wonder if we can make it expand again.

edvin:
I think the parameter to the autocomplete filter should be `comboBox.converter.toString(current)` instead of just `text`.

edvin:
I'll make sure that the data list is updated if the underlying `comboBox.items` property changes.

edvin:
I'm unsure if it's a good idea to change the underlying `comboBox.items` as a result of the filtering. I think users would expect it to be the same instance they passed in, so this could lead to strange bugs. I wonder if we can filter the selection some other way.

aymen083:
I agree with the for the list. It will be more optimized this way.
For the issue the simplest way to solve it is to make a show hide after the combobox.items modifcation. I will see the implementation of the popup if there is a way to force the recalculation.

If we create a separate filtred list we have to modify the popupcontrol to use this new list when available.

Or we can expose the orignal list in a another method.

aymen083:
I made some modification to the implementation of the default filter to not return a new list at evry call using a filtredList only the predicate is updated.
Then the original list can be retrived by the attribute "source" off the filtredList.

And i solved the issue of popup using reflection with a fall to showhide solution.

This is the patch with modification : 
[Rework_of_filtred_list_to_use_FiltredList_and_modify_the_predicate_instead_of_returning_a_.txt](https://github.com/edvin/tornadofx/files/788770/Rework_of_filtred_list_to_use_FiltredList_and_modify_the_predicate_instead_of_returning_a_.txt)

I can make a pull request if you want.

aymen083:
I tried to make the filter not using `combobox.items` but it sems to be complicated and error prone.
To be able to use an other list than `combobox.items` we should reimplement the `com.sun.javafx.scene.control.skin.ComboBoxListViewSkin` using the filtred list instead of `combobox.items` i tried to make it by overriding some methods but i didn't figure how to do it.

With the proposed implementation in the patch of previous comment, using the default filter the user can retrive the original list in the `source` attribute of the filtredList :

Javadoc of FiltredList :
> public final class FilteredList<E> extends TransformationList<E,E>
> Wraps an ObservableList and filters it's content using the provided Predicate. All changes in the ObservableList are propagated immediately to the FilteredList.  

And i the user make modfication in the original list the changes will be propagated to the filtred list. 

What do thing ?

edvin:
I've tried a lot of different approaches myself, and I've reached the same conclusion: This will require a custom Skin to avoid modifications to the source list. I was able to get it working by overriding ComboBoxListViewSkin, but the approach had problems. A complete reimplementation would be best. It's kind of unfortunate, because we'd inherit some maintenance responsibilities for the ComboBox. However, a custom skin could be the gateway to even cooler functionality in the future.

I could try to create a custom skin the coming weekend. If it works, this will be the best approach I think!

edvin:
Let me know when you have time to check the new skin implementation :)

Issue #271, opened 2017-03-13
Non-null parameter being called with null value from tornadofx code


CAD97:
[`Forms.kt:244`](https://github.com/edvin/tornadofx/blob/master/src/main/java/tornadofx/Forms.kt#L244) can run `text.isNotBlank()` while `text: String!` is `null`.

<details><summary><strong>Stack Trace</strong></summary><pre>
SEVERE: Uncaught error
java.lang.IllegalArgumentException: Parameter specified as non-null is null: method kotlin.text.StringsKt__StringsJVMKt.isBlank, parameter $receiver
	at kotlin.text.StringsKt__StringsJVMKt.isBlank(StringsJVM.kt)
	at tornadofx.AbstractField.computePrefHeight(Forms.kt:244)
	at tornadofx.AbstractField.computeMinHeight(Forms.kt:272)
	at javafx.scene.Parent.minHeight(Parent.java:957)
	at javafx.scene.layout.Region.minHeight(Region.java:1401)
	at javafx.scene.layout.Region.computeChildMinAreaHeight(Region.java:1697)
	at javafx.scene.layout.VBox.getAreaHeights(VBox.java:436)
	at javafx.scene.layout.VBox.computeContentHeight(VBox.java:518)
	at javafx.scene.layout.VBox.computeMinHeight(VBox.java:396)
	at javafx.scene.Parent.minHeight(Parent.java:957)
	at javafx.scene.layout.Region.minHeight(Region.java:1401)
	at javafx.scene.layout.Region.computeChildMinAreaHeight(Region.java:1697)
	at javafx.scene.layout.VBox.getAreaHeights(VBox.java:436)
	at javafx.scene.layout.VBox.computeContentHeight(VBox.java:518)
	at javafx.scene.layout.VBox.computeMinHeight(VBox.java:396)
	at javafx.scene.Parent.minHeight(Parent.java:957)
	at javafx.scene.layout.Region.minHeight(Region.java:1401)
	at javafx.scene.layout.Region.computeChildMinAreaHeight(Region.java:1697)
	at javafx.scene.layout.BorderPane.getAreaHeight(BorderPane.java:618)
	at javafx.scene.layout.BorderPane.computeMinHeight(BorderPane.java:415)
	at javafx.scene.Parent.minHeight(Parent.java:957)
	at javafx.scene.layout.Region.minHeight(Region.java:1401)
	at javafx.scene.layout.Region.boundedNodeSizeWithBias(Region.java:1917)
[...]

PR #272, opened 2017-03-13
Tentative fix for edvin/tornadofx/issues/271

CAD97:
As described in https://github.com/edvin/tornadofx/issues/271, the `text` property('s value) on `AbstractField` can sometimes be `null`. This merely makes that explicit and replaces the null-unsafe `text.isNotEmpty()` with the null-safe `!text.isNullOrBlank()`.

I do not have the wherewithal to test this at this time, so request that testing is done upstream before merging this to make sure this doesn't break anything (as unlikely as that is) and to determine if this issue is present elsewhere.


edvin:
Thank you for the thorough report and PR. You are completely right, as long as the text property is defined as nullable, we must treat it as such. Great work!

PR #286, opened 2017-03-26
Add Node event handler for long and short presses

bvolkmer:
This is my first working try.

Used in a datagrid like:
```
selectionModel.selectedIndices.addListener(ListChangeListener {
    while (it.next()) {
        if (it.wasRemoved() && it.list.size <= 0) inSelectionMode = false
}})
cellCache {
    stackpane {
        addShortLongPressHandler(Duration(1000.0), true, { event ->
            if (inSelectionMode) {
                val index: Int = this@datagrid.items.indexOf(it)
                if (selectionModel.isSelected(index)) selectionModel.clearSelection(index)
                else selectionModel.select(it)
            } else {
                controller.open(it)
            }
        }, { event ->
            inSelectionMode = true
            val index: Int = this@datagrid.items.indexOf(it)
            if (selectionModel.isSelected(index)) selectionModel.clearSelection(index)
            else selectionModel.select(it)
        })
    }
}
```

Some side effects like canceling when move off the node before releasing are not considered by now.
A global `var inSelectionMode: Boolean` is used for toggling the toolbar.

### DIFF
--- a/src/main/java/tornadofx/Nodes.kt
+++ b/src/main/java/tornadofx/Nodes.kt
@@ -3,6 +3,7 @@
 package tornadofx
 
 import com.sun.javafx.scene.control.skin.TableColumnHeader
+import javafx.animation.PauseTransition
 import javafx.application.Platform
 import javafx.beans.binding.BooleanBinding
 import javafx.beans.property.DoubleProperty
@@ -33,6 +34,7 @@ import javafx.scene.paint.Paint
 import javafx.stage.Modality
 import javafx.stage.Stage
 import javafx.util.Callback
+import javafx.util.Duration
 import javafx.util.StringConverter
 import javafx.util.converter.*
 import tornadofx.osgi.OSGIConsole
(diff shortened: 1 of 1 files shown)


edvin:
Thank you. The name is not a good fit for TornadoFX, and I wonder if there should be separate functions like `longPress` and `shortPress`? I have a Surface Pro in order, will investigate more when it arrives :)

edvin:
I finally got my Surface Pro yesterday, and started looking at this. Will get back to you soon.

edvin:
Shouldn't this listen to `TouchEvent.TOUCH_PRESSED` and `TouchEvent.TOUCH_RELEASED` instead of `MouseEvent`?

bvolkmer:
Depends in if you want it to work with your mouse, too. Might be better. I'll test it this evening.

edvin:
Hey! I have taken your code and altered it slightly so you can add either short or long press, or both. This improves the syntax and is more flexible. I guess it is bad form to close your PR and implement it myself with your code so I just want to make sure you get all the credit for this. Nice work!

The API is now:

```kotlin
button("Short or longpress me") {
    longpress { println("Long") }
    shortpress { println("Short") }
}
```
They also support the `consume` parameter, and `longpress` supports `threshold` so you can write:

```kotlin
longpress(1.second, true) { println("Long, consumed after one second") }
```
Thank you for your work on this. This feature will make it into the upcoming 1.7.2 release, which is right around the corner :)

bvolkmer:
As it was quite unfinished at the time of posting, it would have been better put in a gist. But thanks anyway. Its nice to see it in the release so soon.

PR #288, opened 2017-03-30
Updated all testapps to be in line with the new action builder.

nimakro:


src/test/kotlin/tornadofx/testapps/GotoViewApp.kt
```diff
@@ -10,11 +10,13 @@ class ContainerView : View("ContainerView") {
 
     override val root = borderpane {
         top {
-            button("Switch view").setOnAction {
-                if (center.lookup("#view1") != null)
-                    subView1.replaceWith(subView2, ViewTransition.Slide(0.2.seconds))
-                else
-                    subView2.replaceWith(subView1, ViewTransition.Slide(0.2.seconds, ViewTransition.Direction.RIGHT))
+            button("Switch view") {
```
t-boom: This could have stayed as `button("Switch View").acction { ... }`

src/test/kotlin/tornadofx/testapps/SwitchViewApp.kt
```diff
@@ -11,11 +11,13 @@ class GotoContainerView : View("ContainerView") {
 
     override val root = borderpane {
         top {
-            button("Goto view").setOnAction {
-                if (center.lookup("#view1") != null)
-                    gotoSubView1.goto(gotoSubView2)
-                else
-                    gotoSubView2.goto(gotoSubView1)
+            button("Goto view") {
```
t-boom: Same here

edvin: Sweet, I merged and fixed those and incorporated even more small syntax improvements.

src/test/kotlin/tornadofx/testapps/GotoViewApp.kt
nimakro: Yes you are right, but is there a diffrence between `button("Switch View").action { ... }` and 
```kotlin
button("...")  {
      action {
          ...
     }
}
```
or why would you prefer the one over the other (saving LOC) ?

t-boom: In this case (with `action` being the only call in the block) there is no difference. I'm not so bothered by lines of code, I just think it's cleaner.

t-boom: My main point was simply that the only change needed was to change `setOnAction` to `action` like you did everywhere else.

edvin: I usually do the same. If there is just a single statement inside a builder, I append it instead of adding a code block. Then I refactor it if I need more statements later.

nimakro: Okay that makes sense to me, but then we should change the other test to be consistent, like in `EventBusTest.kt`
where we have the following code:
```kotlin
     button("Load data") {
             action {
                  fire(GiveMeData)
              }
          }
```
and change it to
```kotlin
     button("Load data").action {
                  fire(GiveMeData)
     }
```
I will go through the tests and make them consistent.

edvin:
Great, thank you :)

PR #292, opened 2017-04-04
Config helper rework

nimakro:
This is my current idea for the config helper. I didn't add the changes to the changelog yet since I am not sure if you accept this implementation. If you do let me know and I will add the information to the changelog. If this will get merged I will also change the guide.

The old implementation allowed to store null values, which the new functions do not allow. Except for the ones from Properties them selfs.
If a read function can't find a given key it will return null. I also introduced 
a `orElse` function to work nicely with nulls which can be called like so: ` string(key).orElse(default)` or you do `string(key) ?: default`

The implementation now allows storing `GLOBAL` and `COMPONENT` configs.
```kotlin
class MyController() : Controller() {
    // You can override the directory per component if you like
    // If the configPath is relative it will be placed as a subfolder 
    // of the configBasePath defined in App. 
    // The Default is null, which means the basePath is used 
    // (I did that to be compatible with the old implementation.
    override val configPath = Paths.get("custom-dir")

    init {
        config {
            set("username" to "user")
        }
        config(GLOBAL) {
            // I added orElse to work with defaults.
            val port = string("port").orElse(7777)
        }
    }
}

class MyController2() : Controller() {
    init {
        config(GLOBAL) {
            val port = string("port").orElse(7777)
        }
    }
}


class Application : App() {
    // The default is Paths.get("conf") to be consitent with the old implementation.
    override val configBasePath = Paths.get("custom-dir")
    // The defualt is null, that means it ill be stored in configBasePath
    override val configPath = Paths.get("custom-dir")

    init {
        // Saving to the global space
        config {
            set("title" to "title")
        }
    }
}
```

### DIFF
--- a/src/main/java/tornadofx/App.kt
+++ b/src/main/java/tornadofx/App.kt
@@ -12,6 +12,8 @@ import java.awt.event.MouseEvent
 import java.awt.event.MouseEvent.BUTTON1
 import java.awt.image.BufferedImage
 import java.io.InputStream
+import java.nio.file.Path
+import java.nio.file.Paths
 import java.util.*
 import javax.imageio.ImageIO
 import javax.swing.SwingUtilities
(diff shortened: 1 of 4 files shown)


edvin:
Thank you for all your valuable input and ideas on this feature. As discussed on Slack, I have implemented these ideas with a slightly different approach.

## Configurable config path

The `App` class can override the default path for config files by overriding `configBasePath`.

```kotlin
class MyApp : App(WelcomeView::class) {
    override val configBasePath: Paths.get("/etc/myapp/conf")
}
```

The path can also be relative, which means the path will be created inside the current working directory. By default, the base path is `conf`.

## Override config path per component

By default, a file called `viewClass.properties` is created inside the `configBasePath`. This can be overriden per component:

```kotlin
class MyView : View() {
    override val configPath = Paths.get("some/other/path/myview.properties")
```

You can also create the View spesific config file below the `configBasePath`, which would make sense in most situations. You do this by accessing the App class through the `app` property of the View.

```kotlin
class MyView : View() {
    override val configPath = app.configBasePath.resolve("myview.properties")
```

## Global application config

The App class also has a `config` property and a corresponding `configPath` property. By default, the configuration for the app class is named `app.config`. This can be overridden the same way you do for a View config.

The global configuration can be accessed by any component at any time in the life cycle of the application. Simply access `app.config` from anywhere to read or write your global configuration.

## JSON configuration settings

The `config` object supports `JsonObject`, `JsonArray` and `JsonModel`. You set them using `config.set("key" to value)` and retrieve them using `config.jsonObject("key")`, `config.jsonArray("key")` and `config.jsonModel("key")`.

What do you think? Can we work with this - do you need more features?

nimakro:
This looks very good and clean I think we can leave it like this for now

edvin:
Great! I will add it to the docs :)

PR #293, opened 2017-04-05
Add Autocomplete Tooltip mode for non editable combobox.

aymen083:
Extract logic about event for autocomplete key event in a separate class
Create an common interface for TextFied and Tooltip Based filter


aymen083:
This is cleaner pull request for Tooltip based combobox autocomplete.

edvin:
Fantastic work, this turned out very nice!

PR #300, opened 2017-04-14
Fix JsonStructure.save(output: Path, vararg options: OpenOption)

Nandi:
Currently the JsonStructure.save(output: Path, vararg options: OpenOption) function just creates an output stream with the path given, but never actually saves the JsonStructure.


edvin:
Ouch, sorry about that - and thank you!

PR #303, opened 2017-04-19
Added lazyContextmenu to create context menus on demand.

dominiks:
The current `contextmenu` function creates the menu immediately but I sometime need them to be created on demand (many items with menus or children change depending on application state) so I added this functionality.

Could also be done with a lazy flag on the existing `contextmenu` function but I think this keeps the functions a bit cleaner and has the same name format as `TreeView<T>.lazyPopulate`.

### DIFF
--- a/src/main/java/tornadofx/Menu.kt
+++ b/src/main/java/tornadofx/Menu.kt
@@ -200,6 +200,7 @@ fun EventTarget.contextmenu(op: (ContextMenu.() -> Unit)? = null): EventTarget {
     } else if (this is Node) {
         this.setOnContextMenuRequested { event ->
             menu.show(this@contextmenu, event.screenX, event.screenY)
+            event.consume()
         }
     }
     return this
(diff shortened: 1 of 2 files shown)


dominiks:
Also fixed an issue with event consumption.

edvin:
Fantastic. Love it :)

src/main/java/tornadofx/Menu.kt
```diff
@@ -200,6 +200,26 @@ fun EventTarget.contextmenu(op: (ContextMenu.() -> Unit)? = null): EventTarget {
...
+}
+
+/**
+ * Add a context menu to the target which will be created on demand.
+ */
+fun EventTarget.lazyContextmenu(op: (ContextMenu.() -> Unit)? = null): EventTarget {
+    var currentMenu: ContextMenu? = null
+    if (this is Node) {
+        this.setOnContextMenuRequested { event ->
+            currentMenu?.hide()
```
t-boom: Won't `currentMenu` always be `null` here? What is the purpose of this call?

PR #304, opened 2017-04-19
Brought some order to the changelog

dominiks:
I suggest we categorize the changes in the changelog a bit to make it easier to spot changes and new features that might necessitate changes in the applications using tornadofx. (The changelog for 1.7.1 was crazy long - which is great - but also hard to parse ;) )

It might also be interesting to highlight completely new features such as the wizard or the autocomplete.


edvin:
Thank you! Let's use this structure going forward.

PR #307, opened 2017-04-21
Added optional orientation parameter to splitpane builder.

dominiks:
Small improvement for adding splitpanes.


edvin:
Thank you :)

PR #315, opened 2017-05-03
Controls:

nimakro:
- Small improvement the label builder is now able to take a graphic node.

@edvin I am not sure if I didn't the label builder with the graphicProperty right (still learning kotlin).


edvin:
Perfect!

PR #339, opened 2017-05-29
Allow to manually unsubscribe from event outside of its handler

AlexP11223:


edvin:
Maybe the registration should have an unsubscribe function on itself?

edvin:
Perfect!

PR #348, opened 2017-06-15
Added extension functions to NumberProperty classes

TheJhonny007:
Added extension functions to NumberProperty classes. They behave like normal Numbers.


edvin:
Fantastic work! Very nice addition to the framework.

PR #351, opened 2017-06-22
master

TheJhonny007:
- Property.plus(), minus(), etc... now return Bindings instead of Properties

- Added Extensions for StringProperty and BooleanPropert

### DIFF
--- a/src/main/java/tornadofx/Properties.kt
+++ b/src/main/java/tornadofx/Properties.kt
@@ -229,46 +229,46 @@ operator fun BooleanProperty.setValue(thisRef: Any, property: KProperty<*>, valu
 operator fun DoubleExpression.plus(other: Number): DoubleBinding = add(other.toDouble())
 operator fun DoubleExpression.plus(other: ObservableNumberValue): DoubleBinding = add(other)
 
-operator fun DoubleProperty.plusAssign(other: Number) = set(get() + other.toDouble())
-operator fun DoubleProperty.plusAssign(other: ObservableNumberValue) = set(get() + other.doubleValue())
+operator fun DoubleProperty.plusAssign(other: Number) { value += other.toDouble() }
+operator fun DoubleProperty.plusAssign(other: ObservableNumberValue) { value += other.doubleValue() }
 
 operator fun DoubleProperty.inc(): DoubleProperty {
-    set(get() + 1)
+    value++
     return this
 }
 
 operator fun DoubleExpression.minus(other: Number): DoubleBinding = subtract(other.toDouble())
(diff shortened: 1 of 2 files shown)


edvin:
This sounds like a good idea, and even though we break the API I think it should be done. It will take me a little while to go through the PR though, will get back to you ASAP.

TheJhonny007:
Great, I was defenitly too fast with the first version, because it was a terrible idea to return Properties instead of Bindings

edvin:
Hehe, yeah sorry I didn't catch that :)

edvin:
It seems like you have double the indentation in the files, so the diff is pretty big. Can you revert the spacing changes so the diff only contains code changes?

TheJhonny007:
Sorry didn't notice you use spaces for indentation and IntelliJ automatically replaced them with tabs when I auto formatted my code

src/main/java/tornadofx/Properties.kt
```diff
@@ -225,89 +225,50 @@ operator fun IntegerProperty.setValue(thisRef: Any, property: KProperty<*>, valu
...
-        = set(get() + other.doubleValue())
+operator fun DoubleExpression.plus(other: Number): DoubleBinding = add(other.toDouble())
+operator fun DoubleExpression.plus(other: ObservableNumberValue): DoubleBinding = add(other)
 
+operator fun DoubleProperty.plusAssign(other: Number) = set(get() + other.toDouble())
+operator fun DoubleProperty.plusAssign(other: ObservableNumberValue) = set(get() + other.doubleValue())
 
 operator fun DoubleProperty.inc(): DoubleProperty {
-    set(get() + 1.0)
+    set(get() + 1)
```
t-boom: This might be clearer as `value += 1.0`, but that may just be me.

t-boom:
There are a number of places where you have code like `set(get() + inc)`. Personally I think it would be clearer to use `value += inc`. but I'm curious to know if I'm the only one.

(This also goes for other operators, e.g. `set(get() - dec)` => `value -= dec`, `set(get() * mul)` => `value *= mul`, etc.)

edvin:
Yeah, it even seems `value` would be slightly more correct if I understand the semantics correctly.

TheJhonny007:
That is only possible if you convert the functions to functions with bodies, since only expressions and no assignments are allowed inline. Otherwise you get this error:

`Error:(r, c) Kotlin: Assignments are not expressions, and only expressions are allowed in this context`

I could expand all those functions, but that would bloat the API.

edvin:
Thank you for this :) Merging now.

PR #370, opened 2017-07-19
changes implementation of param, so nullableParam can be deprecated.

### LINKED ISSUE #369: nullableParam fix?
To start, I read about nullableParam and tried something out.
If the "quirk" in Kotlin was that the defaultValue could not be returned,  if the parameter is nullable, the following code fixes the problem.

```kotlin   
 inline fun <reified T> Component.param2(defaultValue: T? = null): ReadOnlyProperty<Component, T> = 
    object : ReadOnlyProperty<Component, T> {
    override fun getValue(thisRef: Component, property: KProperty<*>): T {
        val nullable = property.returnType.isMarkedNullable
        val param = thisRef.params[property.name] as? T
        if (param == null) {
            if (defaultValue!=null) return defaultValue
            if (nullable) return defaultValue as T
            throw IllegalStateException("param for name [$property.name] has not been set")
        } else {
            return param
        }
    }
}
```

If not, what exactly is the quirk?

tieskedh:
closes #369

### DIFF
--- a/src/main/java/tornadofx/Component.kt
+++ b/src/main/java/tornadofx/Component.kt
@@ -157,21 +157,18 @@ abstract class Component : Configurable {
         override fun getValue(thisRef: Component, property: KProperty<*>): T {
             val param = thisRef.params[property.name] as? T
             if (param == null) {
-                if (defaultValue == null) {
-                    throw IllegalStateException("param for name [$property.name] has not been set")
-                }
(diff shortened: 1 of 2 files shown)


edvin:
Great work, thanks alot!

tieskedh:
you're welcome. 
You're responding quick!!!

Issue #375, opened 2017-07-23
EventTarget.imageview called with ObservableValue<String> containing null produces NullPointerException


DaanVandenBosch:
When you call imageview with a string that's null, it simply doesn't create an Image. Could the same behavior be implemented when you call it with an ObservableValue<String>?

I currently use my own version of imageview which seems to work:

```
fun EventTarget.imageview(url: ObservableValue<String>, lazyload: Boolean = true, op: (ImageView.() -> Unit)? = null)
        = opcr(this, ImageView().apply { imageProperty().bind(objectBinding(url) { value?.let { Image(it, lazyload) } }) }, op)
```

edvin:
Absolutely :) Would you like to submit a PR? If you don't have time, I'll add it.

DaanVandenBosch:
Done. I added a simple unit test too, but I'm still new to JavaFX and TornadoFX so please tell me if the test doesn't really make sense.

edvin:
The test can't hurt :) Can we close the issue?

DaanVandenBosch:
Great, I'll close the issue. And thank you for creating and maintaining such a nice library! Whenever I compare JavaFX examples with TornadoFX examples, the latter ones look much simpler and clearer. I really like how everything is type safe, how easy it is to understand view hierarchies in code and how TornadoFX has so many convenience classes and functions while remaining just a thin layer over JavaFX. The documentation is also superb, I barely even need the JavaFX docs even though I'm completely new to JavaFX. Kudos.

edvin:
Thank you for your feedback, that means a lot :))

PR #377, opened 2017-07-24
EventTarget.imageview now accepts null from an ObservableValue<String>. #375

DaanVandenBosch:


edvin:
This is perfect, thanks! I will update the CHANGELOG as well.

PR #378, opened 2017-07-25
Added Paint.asBackground extension

PvdBerg1998:


t-boom:
I'm not sure if it's worth it to create a new `Paint.kt` file for one extension function. Is there no good place to add it?

PvdBerg1998:
You tell me, haha. I like to add extensions for class X to X.kt. If you know a better destination, let me know, I'll update.

t-boom:
@edvin May have an answer. I believe there is a penalty for each additional file in Kotlin, but I'm not sure how much of an issue it is.

t-boom:
I like the function though. It would be super useful for me.

edvin:
The penalty is basically one extra class, but I agree with @t-boom - we shouldn't add a new file for a single extension function. Maybe we could place it in CSS.kt?

edvin:
Great! It can be moved later without affecting the API so lets not fret about it :) Thanks for this!

PR #379, opened 2017-07-25
Events in fxml controls

TresCe:
Can a method be assigned to an event in fxml without a listener in the initialize?

### DIFF
--- a/src/main/java/tornadofx/Animation.kt
+++ b/src/main/java/tornadofx/Animation.kt
@@ -638,14 +638,20 @@ abstract class ViewTransition {
     /**
      * A [ViewTransition] that can be reversed
      */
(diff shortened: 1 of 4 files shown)


bekwam:
Yes. For example, this FXML adds a delete action to button

            <Button fx:id="deleteButton" mnemonicParsing="false" onAction="#delete">
                <graphic>
                    <FontAwesomeIconView glyphName="CLOSE" size="1.5em" styleClass="close_icon"/>
                </graphic>
            </Button>
It's hooked-in to TorandoFX through a simple function.  No special annotations are needed.

    fun delete() {  // in use; called from .fxml
        if( item != null ) {
            model.remove( item!! )
        }
    }

TresCe:
Thank you! <3 I'll start testing :3

PR #381, opened 2017-07-25
Added onDock javadoc

PvdBerg1998:
Added javadoc to onDock since I was confused about what it was for.
Description as given by @t-boom .


PvdBerg1998:
Woops

PR #384, opened 2017-07-26
Added onDock javadoc

PvdBerg1998:
Added javadoc to onDock since I was confused about what it was for.
Description as given by @t-boom .

Recreating this PR because I accidentally removed the original :)


edvin:
The onDock callbacks are called either when the component becomes the scene root, or when it's root node is attached to another component (using add). Could we reflect that as well in the comment?

PvdBerg1998:
/**
     * Called when a Component when it becomes the Scene root or
     * when it's root node is attached to another Component.
     * @see UIComponent.add
     */

Something like this?

edvin:
That sounds good :)

PvdBerg1998:
So many typo's though

PR #392, opened 2017-07-28
Change scene references to weak references, fix #391

### LINKED ISSUE #391: Memoryleak with FX.Companion.applyStylesheetsTo()
The `stylesheets` ObservableList in FX.kt has `ListChangeListener`s that hold strong references to the `Scene`. This causes memory leaks because the `Scene`s and their content will not be garbage collected anymore.

Screenshot from VisualVM heap dump analyzer:

screenshot

HaasJona:


edvin:
This should be fine. The Scene will persist as long as it's used, so I don't foresee any negative impacts of this fix. Very nice and clean solution. Thank you! PS: You're welcome to send PRs for the CHANGELOG as well when you improve then framework :)

edvin:
We could create a nicer way to work with weak references, so we could write this code as:

```kotlin
private class MyListChangeListener(scene: Scene) : ListChangeListener<String> {
    val sceneRef by weak(scene) { stylesheets.removeListener(this) }

    override fun onChanged(change: ListChangeListener.Change<out String>) {
        sceneRef.ifActive {
            while (change.next()) {
                if (change.wasAdded()) change.addedSubList.forEach { stylesheets.add(it) }
                if (change.wasRemoved()) change.removed.forEach { stylesheets.remove(it) }
            }
        }
    }
}
```

The implementation would probably be something like:

```kotlin
fun <T> weak(referent: T, deinit: (() -> Unit)? = null): WeakDelegate<T> = WeakDelegate(referent, deinit)

class WeakDelegate<T>(referent: T, deinit: (() -> Unit)? = null) : ReadOnlyProperty<Any, DeregisteringWeakReference<T>> {
    private val weakRef = DeregisteringWeakReference(referent, deinit)
    override fun getValue(thisRef: Any, property: KProperty<*>) = weakRef
}

class DeregisteringWeakReference<T>(referent: T, val deinit: (() -> Unit)? = null) : WeakReference<T>(referent) {
    fun ifActive(op: T.() -> Unit) {
        val ref = get()
        if (ref != null) op(ref) else deinit?.invoke()
    }
}
```

Would that make sense?

HaasJona:
It's probably not bad, but the cleaner way would be to use a `ReferenceQueue` so you can clean up collected references without needing to call get to test() if they're still alive. But either should do for now.

edvin:
Cool, I didn't know about that! We should go all the way once and for all, that makes it even nicer to use :)

edvin:
That would entail running a thread and inspecting the references in the queue for every object using this construct, wouldn't? That seems cumbersome.

HaasJona:
Yes, that's why I didn't use this solution. But with not using a cleanup thread, the listeners will still be attached unless an update happens, which causes a smaller memory leak. Which is no problem with scenes as there aren't that many, but if you propose a general solution, I would have used the ReferenceQueue there. You can use one thread and one ReferenceQueue for the whole application, so the overhead isn't that bad.

edvin:
That's right. I think we should use this approach with a cleanup thread, the code shouldn't be very complex, and we have nice entry points for stopping/starting such a thread. I think it's worth it. I don't think this removes the need for checking if the reference is available though, as an object could be garbage collected and attempted accessed before the cleanup thread gets to it, right?

Issue #388, opened 2017-07-27
Make StringConverter parameters contravariant where possible


DaanVandenBosch:
The example below won't compile without the slightly modified converter function. In the current TornadoFX version the StringConverter's type parameter has to be S instead of any super class of S and the TableColumn has to be a TableColumn<T, S?>. So the only type that would work for amount is Number? (nullable).

Current converter method:
```kotlin
fun <T, S : Any> TableColumn<T, S?>.converter(converter: StringConverter<S>): TableColumn<T, S?> = apply {
    cellFormat { text = converter.toString(it) }
}
```

Example with improved converter method:
```kotlin
import javafx.collections.FXCollections
import javafx.scene.control.TableColumn
import javafx.util.StringConverter
import javafx.util.converter.NumberStringConverter
import tornadofx.*

data class Thing(val amount: Int)

class ThingsView : View() {
    private val things = FXCollections.observableArrayList<Thing>()

    override val root = tableview(things) {
        column("Amount", Thing::amount) {
            converter(NumberStringConverter())
        }
    }
}

fun <T, S> TableColumn<T, S>.converter(converter: StringConverter<in S>): TableColumn<T, S> = apply {
    cellFormat { text = converter.toString(it) }
}
```

Is this a good solution and are there other places where we can apply this (list views, labels)? Is there a reason why S has an explicit upper bound on Any, isn't this implicitly the case?

As for labels, is it somehow possible to bind a label to a SimpleObjectProperty<Thing> and have it show the thing's amount (formatted by a NumberStringConverter of course) by passing a property reference to the label factory function, similar to how column works in table views?

I'll open a pull request as soon as I get more info and no body points out any flaws or regressions.

DaanVandenBosch:
Would this make sense for labels? It would be nice to reuse converters between e.g. master and detail views where you want to format things the same way.

```kotlin
label(amount, NumberStringConverter())

inline fun <reified T> EventTarget.label(observable: ObservableValue<T>,
                                         converter: StringConverter<in T>,
                                         graphicProperty: ObjectProperty<Node>? = null,
                                         noinline op: (Label.() -> Unit)? = null) = label().apply {
    textProperty().bind(observable.stringBinding { converter.toString(it) })
    if (graphic != null) graphicProperty().bind(graphicProperty)
    op?.invoke(this)
}
```

DaanVandenBosch:
I finally gave it a shot at #398.

edvin:
[...]

PR #398, opened 2017-08-05
StringConverter parameters are now contravariant where possible, and EventTarget.label now accepts an optional StringConverter argument. #388

DaanVandenBosch:


edvin:
Very nice work! All tests pass, and I've tested it against several of my larger applications with no issues.

PR #402, opened 2017-08-08
Added generic properties selectedCell, selectedColumn, and selectedValue for TableView and TreeTableView

gbui:


edvin:
Neat, I like it :))

PR #403, opened 2017-08-08
Fixed AutoCompleteComboBoxSkin preferred width binding

PvdBerg1998:


PvdBerg1998:
Still discussing about this, do not merge yet

PvdBerg1998:
Looking good!

edvin:
Nice! But maybe we can skip "use" in "useAutomaticPopupWidth"?

PR #406, opened 2017-08-12
Tidied up animation chain helpers and added point(x, y)

gbui:
(I wanted to convert `xy` infix calls to `point(x, y)` as I think `x xy y` is a bit hard to read, but I'll wait for your feedback before I do that.)

### DIFF
--- a/src/main/java/tornadofx/Animation.kt
+++ b/src/main/java/tornadofx/Animation.kt
@@ -294,11 +294,26 @@ fun Node.transform(time: Duration, destination: Point2D, angle: Number, scale: P
  * @param op Modify the animation after it is created
  * @return A ParallelTransition
  */
-fun Animation.and(vararg animation: Animation, op: (ParallelTransition.() -> Unit)? = null)
-        = ParallelTransition(this, *animation).apply { op?.invoke(this) }
+fun Animation.and(vararg animation: Animation, op: (ParallelTransition.() -> Unit)? = null) : ParallelTransition {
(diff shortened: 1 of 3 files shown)


edvin:
Yeah, I agree. `xy` was a bit vague. @t-boom What do you think? Maybe I was wrong to insist on `xy`?

t-boom:
I tend to avoid infix functions, so I'd use `point`, but there is a pretty good personal bias there.

@gbui You may also want to add a `point(x: Number, y: Number, z: Number): Point3D` as well, though I'm not sure how often TornadoFX users use the 3D features of JavaFX.

Also, your functions could be written a little more concisely if you want (it's purely a matter of personal taste). For example:

```kotlin
infix fun Animation.and(animation: Animation) = if (this is ParallelTransition) apply {
    children += animation
} else ParallelTransition(this, animation)
```

t-boom:
If you do decide to remove the `xy` infix, I'd recommend deprecating it with a `ReplaceWith("point(x, y")`.

edvin:
Thanks for the feedback @t-boom. We could leave the xy infix function but let the framework use the point function instead. Good idea to add a point(x, y, z) as well!

gbui:
Yes, thank you for the feedback.

In regards to writing the infix functions differently, they _can_ be written shorter. However, I think your suggestion is a little unreadable, at least at first glance. Indeed, there are other ways to shorten the function, but I also think that having similarities between the infix and non-infix functions allows for easier understanding.

PR #412, opened 2017-08-15
master

TheJhonny007:
Added I18N support for the required message (english and german to start with)


TheJhonny007:
I think if text is shown to the user it should be internationalized. This was the only occurence of custom text I found. Other locales can be added by people who speak other languages.

edvin:
Nice! I should have done this from the start!

PR #415, opened 2017-08-16
json.add allowing use of Array now

vertexpipeline:


edvin:
I already added this functionality as I wrote on the ticket :)

PR #425, opened 2017-08-26
added option to increment and decrement spinners by scrolling

TheJhonny007:


TheJhonny007:
I could also set the default to true, but that might break existing code.

edvin:
Nice one!

PR #438, opened 2017-09-04
Changed PropertyHolder.fontFamily to MultiValue

gbui:
Used for fallback fonts

Kotlin:
````kotlin
label {
    fontFamily = multi("Helvetica", "Arial", "sans-serif")
}
````

CSS:
````css
.label {
    -fx-font-family: "Helvetica", "Arial", "sans-serif";
}
````


edvin:
As this would be a breaking change, and since the call is still not type safe, I don't think it's worth making this change. It doesn't add much value as opposed to adding a comma separate String.

t-boom:
Good call since the JavaFX documentation specifically says:

> JavaFX CSS does not support comma-separated series of font family names in the -fx-font-family property.

gbui:
Ah, my bad. I'll make sure to check the documentation in the future.

t-boom:
It's all good. I have to go back and check the reference almost every time I'm messing with the code too. There's so many little gotchas between normal CSS and JavaFX CSS to trip you up.

PR #439, opened 2017-09-04
refactoring to make use of require and requireNotNull, where it makes sense

tieskedh:


src/main/java/tornadofx/Component.kt
```diff
@@ -712,57 +712,54 @@ abstract class UIComponent(viewTitle: String? = "", icon: Node? = null) : Compon
...
-                        scene = getRootWrapper().scene
-                        this@UIComponent.properties["tornadofx.scene"] = getRootWrapper().scene
-                    } else {
-                        Scene(getRootWrapper()).apply {
-                            if (escapeClosesWindow) {
-                                addEventFilter(KeyEvent.KEY_PRESSED) {
-                                    if (it.code == KeyCode.ESCAPE)
-                                        close()
-                                }
+            require(getRootWrapper() !is Parent) { "Only Parent Fragments can be opened in a Modal" }
```
t-boom: Shouldn't this be `require(getRootWrapper() is Parent)`? The `!is` seems backwards.

tieskedh: YES

PR #442, opened 2017-09-06
range-checks

tieskedh:
- refactoring: adds rangechecks where possible
- fixes: last item in DataGridFocusModel was not possible


edvin:
Great @tieskedh - another super cleanup job :)

PR #449, opened 2017-09-08
Style the accent color with type-safe CSS, closes #448

### LINKED ISSUE #448: Style the root accent color with type-safe CSS
I'd like to adjust the accent color defined by modena.css using type-safe CSS builder. It seems there was a [prior decision](https://github.com/edvin/tornadofx/issues/197#issuecomment-269014914) to defer that until it was requested.

```kotlin
class CustomView : View("Accent Override") {
  override val root = HBox()
  init {
    with (root) {
      style {
        accentColor = Color.RED
      }
    }
  }
}
```

carltonwhitehead:
Override -fx-accent with type-safe CSS property accentColor


edvin:
Thank you, this looks good :)

PR #450, opened 2017-09-09
switch argument names of box function

kkriske:
In the 2 argument version of the `box` function the behaviour is correct when assuming the same argument order as normal css, but when named parameters are used, `horizontal` and `vertical` are switched.

### DIFF
--- a/src/main/java/tornadofx/CSS.kt
+++ b/src/main/java/tornadofx/CSS.kt
@@ -1271,7 +1271,7 @@ fun <T : Node> T.setId(cssId: CssRule): T {
 // Containers
 
 fun <T> box(all: T) = CssBox(all, all, all, all)
-fun <T> box(vertical: T, horizontal: T) = CssBox(vertical, horizontal, vertical, horizontal)
+fun <T> box(horizontal: T, vertical: T) = CssBox(horizontal, vertical, horizontal, vertical)
 fun <T> box(top: T, right: T, bottom: T, left: T) = CssBox(top, right, bottom, left)
 data class CssBox<out T>(val top: T, val right: T, val bottom: T, val left: T)
 


t-boom:
That's different from what I've read (though I'm no expert). According to W3C:

> If the padding property has two values:
> `padding: 25px 50px;`
> - top and bottom paddings are 25px
> - right and left paddings are 50px

t-boom:
Unfortunately, JavaFX CSS doesn't officially support the shorthand, so it doesn't say for sure there. According to [the documentation](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/doc-files/cssref.html#region), it only accepts `<size> | <size> <size> <size> <size>`, so I based the box function off standard CSS rules.

kkriske:
I too checked W3C so given the same sample:

>If the padding property has two values: `padding: 25px 50px;`
>
>top and bottom paddings are 25px
>right and left paddings are 50px

the first argument applies to top and bottom, which I would argue are the horizontal ones
the way it is in the current code, the argument named `horizontal` applies to the left and right edge

the order of the arguments is correct but the naming is the issue

t-boom:
Ah, so we have a different understanding of horizontal and vertical. I understand them the other way around (obviously since I implemented them that way), so we should get some input from others to see what is the most common understood meaning for the two.

@edvin (and anyone else that wants) Any input?

edvin:
To me, the current approach seems natural. Seeing as people will misunderstand this anyway based on how they interpret it, I don't think it holds much value to change. In hindsight, maybe we should have called the parameters topBottom anf leftRight to avoid confusion, buy that's not very slick. 

I vote that we leave it,  but add documentation to the functions instead.

edvin:
I will close this PR for now. Unless some more compelling arguments come along, we'll leave the current approach. Thanks for your effort @kkriske :)

t-boom:
Agreed. It's always nice to have someone double checking my assumptions. Keep them coming!

PR #452, opened 2017-09-11
Refactoring/files/nodes

tieskedh:
- adapters have more info in generics
- refactoring nodes
- lookup stops after he found the first
- node refactoring
- introduced constants-file


tieskedh:
Now doing this per issue of the design guide

PR #454, opened 2017-09-11
CollectionPropertyDelegates

TheJhonny007:
Added CollectionPropertyDelegates

### DIFF
--- a/src/main/java/tornadofx/ItemControls.kt
+++ b/src/main/java/tornadofx/ItemControls.kt
@@ -38,7 +38,7 @@ import kotlin.reflect.KProperty1
 /**
  * Create a spinner for an arbitrary type. This spinner requires you to configure a value factory, or it will throw an exception.
  */
-fun <T> EventTarget.spinner(editable: Boolean = false, property: Property<T>? = null, op: (Spinner<T>.() -> Unit)? = null): Spinner<T> {
+fun <T> EventTarget.spinner(editable: Boolean = false, property: Property<T>? = null, enableScroll: Boolean = false, op: (Spinner<T>.() -> Unit)? = null): Spinner<T> {
     val spinner = Spinner<T>()
     spinner.isEditable = editable
     opcr(this, spinner, op)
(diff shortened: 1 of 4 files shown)


edvin:
Looks good! I will merge and add to CHANGELOG. If you remember it, please add to the CHANGELOG together with the PR :)

TheJhonny007:
I will do that in the future :D

edvin:
These are causing problems because they don't allow for null. This is problematic since the underlying property might contain null.

TheJhonny007:
Can you give me an example, where it creates a problem?

edvin:
This should illustrate the issue:

```kotlin
class X {
    val listProp = SimpleListProperty<Int>()
    var list by listProp
}

fun main(args: Array<String>) {
    val x = X()
    println(x.list)
}
```

TheJhonny007:
This is actually not that easy to solve... I could specify the return type as nullable, but then you have to carry that through the whole code. 

I also stumbled about another Problem:

```kotlin
class StringHolder {
    val stringProperty = SimpleStringProperty()
    var string: String by stringProperty
}

fun main(args: Array<String>) {
    val x = StringHolder()
    println(x.string)
}
```

`output: null`

That shouldn't happen since intField is declared as not nullable.

edvin:
This is probably because it interprets String as a platform type, and it's actually the desired behavior. Changing this would break a lot of existing apps.

TheJhonny007:
```kotlin
class MutableListProperty<E>(list: MutableList<E> = mutableListOf<E>().observable(), bean: Any? = null, name: String = "") 
    : ListPropertyBase<E>(list as? ObservableList<E> ?: list.observable()) {
    
    private val beanField: Any? = bean
    private val nameField: String = name

    override fun getName() = nameField
    override fun getBean() = beanField

    override fun set(newValue: ObservableList<E>) = super.set(newValue)
    override fun get(): ObservableList<E> = super.get()
}

class MutableListPropertyDelegate<E>(private val listProperty: MutableListProperty<E>) {
    operator fun getValue(thisRef: Any, property: KProperty<*>): MutableList<E> = listProperty.get()
    operator fun setValue(thisRef: Any, property: KProperty<*>, list: MutableList<E>) 
            = listProperty.set(list as? ObservableList<E> ?: list.observable())
}

fun <E> property(listProperty: MutableListProperty<E>) = MutableListPropertyDelegate(listProperty)

class ImmutableListProperty<E>(list: List<E> = emptyList<E>().observable(), bean: Any? = null, name: String = "") 
    : ListPropertyBase<E>(list as? ObservableList<E> ?: list.observable()) {
    
    private val beanField: Any? = bean
    private val nameField: String = name

    override fun getName() = nameField
    override fun getBean() = beanField

    override fun set(newValue: ObservableList<E>) = super.set(newValue)
    override fun get(): ObservableList<E> = super.get()

    operator fun getValue(thisRef: Any, property: KProperty<*>): List<E> = get()
    operator fun setValue(thisRef: Any, property: KProperty<*>, list: List<E>) 
            = set(list as? ObservableList<E> ?: list.observable())
}

class ImmutableListPropertyDelegate<E>(private val listProperty: ImmutableListProperty<E>) {
    operator fun getValue(thisRef: Any, property: KProperty<*>): List<E> = listProperty.get()
    operator fun setValue(thisRef: Any, property: KProperty<*>, list: List<E>) 
            = listProperty.set(list as? ObservableList<E> ?: list.observable())
}

fun <E> property(listProperty: ImmutableListProperty<E>) = ImmutableListPropertyDelegate(listProperty)
```

```kotlin
class MutableListHolder {
    val listProperty: ListProperty<String> = MutableListProperty<String>()
    var list: MutableList<String> by property(listProperty)
}

class ImmutableListHolder {
    val listProperty: ListProperty<String> = ImmutableListProperty<String>()
    var list: List<String> by property(listProperty)
}
```

That would be my best solution. Because ListProperty already has a getValue and setValue method it doesn't let me define the delegate methods inside the class...

TheJhonny007:
Forget about that solution. The super classes are still allowing null and immutableLists as parameters...

PR #456, opened 2017-09-12
Refactoring/simple classes

tieskedh:


tieskedh:
now doing this per issue of the design guide

PR #466, opened 2017-09-24
add imports to helloworld example

mgroth0:


mgroth0:
fixes#465

edvin:
Hehe, thanks :) I see how others might run into this trap as well so I guess it makes sense to add the imports :)

PR #468, opened 2017-09-25
Added `Fragment` support for `TreeCell`

nimakro:
I realized that we are still missing `Fragment`support for `TreeCell` in a `TreeView`, this PR is adding support for it. 
@edvin  Before you merge one quick questions since we have the `TreeView`file now, should I move all other extension functions targeting `TreeView` from `Component`to the `TreeView`file. If so I will push another commit.


edvin:
@nimakro Thanks! And yes, please move :)

PR #471, opened 2017-09-30
remove 'lightweight' from README

ayrat555:
from Talking Kotlin 'SEPTEMBER 28, 2017'


edvin:
I’ve thought about doing that for a while. It even used to say extremely lightweight back in the day :) We’ve got so many features now that lightweight might not be a good description anymore :)

PR #479, opened 2017-10-04
Minor fix in ConfigProperties (invalid returned type)

genobis:
Fixed minor typing issue with ConfigProperties.int() method. Now it returns Int?, as it should, instead of Any? (it happened because of java.lang.Integer usage instead of kotlin's Int).


edvin:
Thank you :)

PR #484, opened 2017-10-05
Fix #481 - re-use StringBuilder when building CSS stylesheet.

### LINKED ISSUE #481: Change `Rendered` to use a StringBuilder instead of returning a String
The implementation of CSS rendering is extremely inefficient.
It creates a lot of Strings separately, then merges them together by appending them one by one, then finally turns the result into a byte-array while at the same time holding on to the original String, causing not only a CPU bottleneck, but a memory bottleneck.

The best solution would be to build the CSS with a single StringBuilder by passing it into each `Renderable` (or whatever the more appropriate name for this interface would be).

CSS stylesheets can grow quite large and numerous for a professional application, and spending hundreds of milliseconds rendering it would make it impossible to use type-safe stylesheets for a snappy app.

renatoathaydes:
Hello again!

I did what I suggested in ticket #481, it was not too hard...

But the result totally surprised me.

screen shot 2017-10-05 at 15 52 16

Looks like it didn't make any difference... it anything, it got a little bit slower. I know from previous experience that using a single builder instead of constantly appending Strings should have made a positive difference, but as always with performance, you can never be sure.

I decided to make a pull request anyway because with further profiling, this may turn out to be good anyway. But feel free to decline it as it did not deliver on the improved performance it was supposed to bring.

Notice that I wrote a test that creates a largish spreadsheet and prints how long it took (that's how I got this chart). I would consider keeping that and using it to make sure that when you refactor the CSS, you don't make it a lot slower... It is already quite sluggish, to be honest... How can a CSS with a few dozen lines take 200ms to render? I would expect an order of magnitude less... but the problem seems to be somewhere else :)

Further profiling is required.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,6 +70,17 @@
             <version>${de.jensd.fontawesomefx.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-stdlib-jre8</artifactId>
+            <version>${kotlin.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-test</artifactId>
+            <version>${kotlin.version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <distributionManagement>
@@ -124,7 +135,7 @@
     </scm>
 
     <build>
-        <testSourceDirectory>${project.basedir}/src/test/kotlin</testSourceDirectory>
+        <testSourceDirectory>src/test/kotlin</testSourceDirectory>
 
         <pluginManagement>
             <plugins>
@@ -152,13 +163,14 @@
                 <executions>
                     <execution>
                         <id>compile</id>
-                        <phase>process-sources</phase>
+                        <phase>compile</phase>
                         <goals>
                             <goal>compile</goal>
                         </goals>
                     </execution>
                     <execution>
                         <id>test-compile</id>
+                        <phase>test-compile</phase>
                         <goals>
                             <goal>test-compile</goal>
                         </goals>
--- a/pom.xml
+++ b/pom.xml
@@ -70,17 +70,6 @@
             <version>${de.jensd.fontawesomefx.version}</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.jetbrains.kotlin</groupId>
-            <artifactId>kotlin-stdlib-jre8</artifactId>
-            <version>${kotlin.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>org.jetbrains.kotlin</groupId>
-            <artifactId>kotlin-test</artifactId>
-            <version>${kotlin.version}</version>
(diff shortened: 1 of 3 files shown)


edvin:
I belive you're on a wild goose chase :) The speed issue here is not the rendering at all. To verify, do the following:

Make a copy of your `largeStylesheetCanBeRenderedWithoutErrors` and name it `primeLargeStylesheetCanBeRenderedWithoutErrors`. Now call the prime function outside of `withTimer`, but inside of `largeStylesheetCanBeRenderedWithoutErrors`. You'll see that the stylesheet inside of `withTimer` renders in just a few milliseconds. On my laptop, about 6ms on average.

This is all with the existing approach.

The issue is simply that there is a lot of initialization happening with different JavaFX constructs as you access them the first time. Your proposed PR also removes the possibility of parallell processing as the different parts of the stylesheet renders. 

The initialization of these JavaFX constructs will have to happen to be able to utilize JavaFX at all, so it's a non issue, and nothing we can do anything about, nor is it a delay introduced by TornadoFX, so I think we're all good :)

renatoathaydes:
I am not so convinced. The JVM is amazing optimising stuff after stuff is run a few times, but initializing an app you only get one shot to render Css, so this one shot has to be fast.

Did you try profiling it to see where the time was spent?

edvin:
I did not yet, so we should try that to see where the bottle neck is. My bet is on initialization happening inside of a JavaFX construct, but that should be possible to confirm :) I actually don't even have a profiler set up on my laptop, will try when I get home unless you beat me to it? :)

edvin:
I was able to determine that the rendering is not the problem though. Just running the code inside init takes about the same time as running the code _and_ rendering, maybe 0-2ms difference.

renatoathaydes:
Strange, I see something else:

```
create-stylesheet (Total time): 46 ms
large-stylesheet (Total time): 319 ms
```

renatoathaydes:
Just to create an instance of `Stylesheet` takes 46ms. That's probably due to so many delegate properties in it?! But the bulk of the time I can see is spent on rendering.

renatoathaydes:
Hm... I assumed rendering was happening within the `stylesheet` method, but it happens later on `shouldEqual`... so you're correct, creating the stylesheet is the slow process, not rendering it.

Later I will benchmark the rendering again that explains why my changes had no effect :)

Anyway, here's where the time was spent per method call in the test example (values are ms from start):

```
START=0, barChart=99, imageView=108, mixin1=131, mixin2=135, box=151, s=176
```

The total was 211ms in this run, so rendering would be around 30 ms or so.

I am going to travel soon so unfortunately I will not have much time to continue with this... I know it's not a critical issue, but I love performance challenges, so I will have a look again when I get time if you guys don't want to worry about this.

edvin:
Thanks for your effort so far. I think you might be onto something with the delegate functions. The stylesheet has quite a few properties created this way, and they all instantiate yet another class. It would be interesting to try to implement them as getters/setters instead. If you find time to work more on this that would be very much appreciated :)

edvin:
Closing this issue as it seems that StringBuilders isn't the solution. If you have time to work on an alternative to all the delegates we should create a new issue. Closing for now :)

PR #488, opened 2017-10-09
Add a customitem menuitem builder.

melston:


melston:
@edvin , I'm still not that familiar with JavaFX so I wasn't sure what was safe to remove or keep from the original `item` extension.  If you are referring to the inclusion of the line:
```kotlin
graphic?.apply { menuItem.graphic = graphic }
```
then I am happy to remove it.  I haven't seen a lot of good material on `CustomMenuItem` so I wasn't sure if that line was appropriate or not.

edvin:
Thank you. As briefly mentioned on Slack, we needed to make some modifications. I've pushed a `customitem` builder now. You use it like this:

```kotlin
customitem {
    vbox {
        text("Some text or other node")
        button("A button") {
            action {
                println("You clicked!")
            }
        }
    }
}
```

Issue #489, opened 2017-10-09
ComboBox button cell formatter doesn't work correctly


genobis:
https://github.com/edvin/tornadofx/blob/957ee443a73c901515677548e3a5ac7ab5ba454e/src/main/java/tornadofx/ItemControls.kt#L130

User-provided formatter is tightly tied to the `ListView` (as a property). Thing is, cell factory is invoked with undefined `ListView` and thus `ListCell` instance created this way has no formatter. This results in blank combox button (at least until selection is made).

```kotlin
    cellFactory = Callback {
        it?.properties?.put("tornadofx.cellFormat", formatter)
        SmartListCell(scope, it)
    }
    if (formatButtonCell) {
        Platform.runLater {
            buttonCell = cellFactory.call(null)
        }
    }
```

The only way around is to disable this feature (by setting `formatButtonCell` to `false`, because it defaults to `true`) and manually set `ComboBox.buttonCell` value.

I was able to fix it on my own in TornadoFX sources, will push it right away.

genobis:
Fix merged.

PR #490, opened 2017-10-09
#489 fixed ComboBox button cell formatter issue

genobis:
SmartListCell now can have (optionally, so this change should break nothing) custom-defined properties.

New field (smartProperties) is introduced. If listView is set at creation time, its properties are used (provided properties, if any, are  added to list properties), otherwise provided properties are copied and safely wrapped in the new instance of observable map.

All references to listView properties are replaced with references to smartProperties.


edvin:
Thanks, that makes sense :) Also merged in the jdk9 branch.

genobis:
Thanks :) I wasn't sure if I'm getting it right (internal or not, it IS an API change after all). It's nice to be helpful.

Issue #491, opened 2017-10-10
Bug: unable to close InternalWindow due to unsafe type casting


genobis:
Method `findParentOfType` (one of the important ones, I suppose) does not work correctly. In consequence, closing `InternalWindow` from within an embedded component is not possible (`close` method invoked directly for window instance works fine, the problem is with getting that instance). I have even written a test for this particular case and a following exception occurs:

```java
java.lang.RuntimeException: java.lang.ClassCastException: javafx.scene.layout.StackPane cannot be cast to tornadofx.InternalWindow

	at org.testfx.util.WaitForAsyncUtils.waitFor(WaitForAsyncUtils.java:161)
	at org.testfx.api.FxToolkit.waitForSetup(FxToolkit.java:269)
	at org.testfx.api.FxToolkit.setupFixture(FxToolkit.java:218)
	at tornadofx.tests.InternalWindowClosingTest.testClosing(InternalWindowClosingTest.kt:26)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
[...]

PR #492, opened 2017-10-10
#491 - fixed Node.findParentOfType + test

genobis:
Fixed for #491 (broken Node.findParentOfType)

`as?` operator does not work correctly when it comes to erasure and in effect any non-null object was returned, which caused `ClassCastException` at later point.

Fortunately `KClass` instance was available, so it was possible to make truly type-aware safe cast.

A test case has been added as well.


edvin:
Nice work, I wasn't aware of that!

PR #496, opened 2017-10-14
Some additonal eyecandy for async tasks

genobis:
A node (any node) can now be covered with customizable overlay until background task is executed.  A new Latch class allows for finer control, including multithreading, timeouts and bindings.

Docs with some basic examples and test included.

----

I'm working on an application built on top of TFX and while `runAsyncWithProgress` is very cool (in fact, inspired me to do this), I missed the ability to mask entire UI (or only part of it) with transluscent overlay. And I have some legacy, callback-based API which doesn't really play along with mechanisms already built into TFX.

This commit fills the gap for me - a new `Node.runAsyncWithOverlay` method allows to block UI, and Latch (which is just a JavaFX friendly extension of `CountDownLatch`) gives some additional control.

### DIFF
--- a/src/main/java/tornadofx/Async.kt
+++ b/src/main/java/tornadofx/Async.kt
@@ -1,0 +1,140 @@
+package tornadofx
+
+import javafx.beans.property.ReadOnlyBooleanProperty
+import javafx.beans.property.ReadOnlyBooleanWrapper
+import javafx.concurrent.Task
+import javafx.scene.Node
+import javafx.scene.layout.BorderPane
+import javafx.util.Duration
+import java.util.concurrent.CountDownLatch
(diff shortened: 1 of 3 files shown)


edvin:
Nice work! Remember to add to the changelog as well :) I will do it now :)

genobis:
@edvin Thanks :) I will next time. Just weren't sure if it's going to be merged at all ;)

Would it be OK if I refactored existing concurrency-related code into `Async.kt` (without breaking API, of course)? I made the new file because that new code (especially `Latch`) didn't really fit anywhere else, but I don't want it to get it scattered everywhere and messy.

edvin:
Absolutely, please do :) Thanks a lot!

PR #497, opened 2017-10-16
Concurrency-related code refactored and moved to Async.kt

genobis:
* Threads are now created and maintained by a cached thread pool
* concurrency-related functions and helpers moved from `Component.kt`, `lib.kt` and `FX.kt` to `Async.kt`
* `runAsync` and `ui` methods can now be called from anywhere
* `runAsync` kept also in UIComponent because of scope usage, so if called from within UIComponent it's behavior will remain unaltered.
* `runAsyncWithProgress` has now a latch supporting variant
* `runAsyncWithOverlay` some minor changes to docs

------

No test this time nor changelog entry - it's mostly refactoring. All existing ones should pass, as there were no API-breaking changes. Just some functions, previously not accessible without a good reason, now more widely available, but I don't think it's worth mentioning at all.

### DIFF
--- a/src/main/java/tornadofx/Async.kt
+++ b/src/main/java/tornadofx/Async.kt
@@ -1,13 +1,169 @@
 package tornadofx
 
-import javafx.beans.property.ReadOnlyBooleanProperty
-import javafx.beans.property.ReadOnlyBooleanWrapper
+import com.sun.javafx.tk.Toolkit
+import javafx.application.Platform
+import javafx.beans.property.*
+import javafx.beans.value.ChangeListener
+import javafx.beans.value.ObservableValue
 import javafx.concurrent.Task
 import javafx.scene.Node
+import javafx.scene.control.Labeled
+import javafx.scene.control.ProgressIndicator
 import javafx.scene.layout.BorderPane
+import javafx.scene.layout.Region
 import javafx.util.Duration
+import java.util.*
 import java.util.concurrent.CountDownLatch
+import java.util.concurrent.Executors
+import java.util.concurrent.ThreadFactory
 import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicLong
+
+internal val tfxThreadPool = Executors.newCachedThreadPool(object : ThreadFactory {
+    private val threadCounter = AtomicLong(0L)
+    override fun newThread(runnable: Runnable?) = Thread(runnable, "tornadofx-thread-${threadCounter.incrementAndGet()}")
+})
+
+fun <T> task(taskStatus: TaskStatus? = null, func: FXTask<*>.() -> T): Task<T> = FXTask(taskStatus, func = func).apply {
+    setOnFailed({ Thread.getDefaultUncaughtExceptionHandler().uncaughtException(Thread.currentThread(), exception) })
+    tfxThreadPool.execute(this)
+}
+
+fun <T> runAsync(status: TaskStatus? = null, func: FXTask<*>.() -> T) = task(status, func)
+
+infix fun <T> Task<T>.ui(func: (T) -> Unit) = success(func)
+
+infix fun <T> Task<T>.success(func: (T) -> Unit) = apply {
+    Platform.runLater {
+        setOnSucceeded { func(value) }
+    }
+}
+
+infix fun <T> Task<T>.fail(func: (Throwable) -> Unit) = apply {
+    Platform.runLater {
+        setOnFailed { func(exception) }
+    }
+}
+
+/**
+ * Run the specified Runnable on the JavaFX Application Thread at some
+ * unspecified time in the future.
+ */
+fun runLater(op: () -> Unit) = Platform.runLater(op)
+
+/**
+ * Run the specified Runnable on the JavaFX Application Thread after a
+ * specified delay.
+ *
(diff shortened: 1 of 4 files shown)


edvin:
Great work, but I think you removed a `runAsync` version from `UIComponent` which included a taskStatus object extracted from the component scope. Was that intentional?

genobis:
@edvin No, it's still there ;) I actually even mentioned this very case in commit message. This is the only piece of concurrency-related code I'm aware of that I left in original place. Not unchanged, though. But you deserve some explanation.

In `Async.kt` I *added* this:

`fun <T> runAsync(status: TaskStatus? = null, func: FXTask<*>.() -> T) = task(status, func)`

This is actually the same as `task` but I believe `runAsync` is more _natural_ in terms of TFX and more widely used (I suppose `task` is used rather internally by TFX, but it's public, so we can never be sure). Anyway, this way it can be invoked from anywhere, including non-components, the difference is that `status` is `null` by default. This is because it's unaware of scope, of course.

Now, in `Component.kt` I changed this:

`fun <T> runAsync(status: TaskStatus? = find(scope), func: FXTask<*>.() -> T) = task(status, func)`

to this:

`fun <T> runAsync(func: FXTask<*>.() -> T) = runAsync(find(scope), func)`

so it's actually still there. If you invoke it from within `UIComponent` and *don't* provide a custom `TaskStatus` (which was the case when `TaskStatus` would be taken from the scope before my changes), it will work as before. If you provide `status` directly, the variant from `Async.kt` would be invoked.

I promised no broken API, remember ;) ?

genobis:
@edvin considering your previous commit (restoring `ui` in `Component.kt`) I suppose I *did* broke after all, despite all my promises. This case is probably very similar. If someone uses `runAsync` from `UIComponent` *and* sets custom `status`, this will require importing `runAsync`.

I'll fix it right away.

edvin:
Hehe, yes, that's right. It's a sneaky kind of API breakage though, but in effect it's the same as moving a function from one package to another, meaning that it requires an additional import, unless you have `tornadofx.*` already.

I kind of regret that I added the version inside Component, but the idea was to increase discoverability, since it doesn't require an import. The Kotlin IDEA plugin is much better than it was initially, so this is a non issue now, but it would be bad to break the API anyways.

genobis:
Well, I *do* have `tornadofx.*` everywhere, and so I lost my guard ;) If (or when) TFX reaches another major version, some major cleanups may be a good idea, but right now I concur - in this case broken API is just a bug.

While I don't feel I'll be able to put all my application logic in components (although I'll try to), I actually like the fact that they have these built-in methods. This discoverability is a good thing, the bad one is that when I got used to them, I couldn't use them anywhere else ;)

Anyway, PR #499 should fix the issue. There were no more API changes like that (extracting methods) in #497, so it should be OK now.

Romanow88:
@genobis hey, thanks for the PR, I've written something similar to this in my project and wanted to do a PR myself but you were first :)

Only thing that I would add - right now there is no way to access the tasks from other components (aside from the task status). Would be nice to be able to access either a task or a limited interface of task methods - for me the scenario is canceling the task from a separate UI component.

edvin:
@romanow88 I saw that you commented about access to the underlying task, but I can't find the comment. Just to clarify, `runAsync` runs in the context of the Task it creates, so `this` is always the Task itself. You can call any function on Task directly inside of `runAsync`.

Romanow88:
@edvin tnx for the answer, I've figured that out while looking trough the code and so I deleted my comment. 

Right now I'm trying to implement canceling a task that is submitted to an executor service (my own impl, not the one submitted by genobis). Seems that just calling Task.cancel(true) does not set the thread interrupt flag to true so current working solution is to store a reference to the FXTask in a ThreadLocal and check for .isCanceled in critical points of the task. 

Any idea how to do that nicely?

P.S. maybe a good idea to consider using the javafx.concurrent.Service to handle tasks as it already have the Executor and some state management.

genobis:
@Romanow88 

From [JavaFX 8 interoperability tutorial](https://docs.oracle.com/javase/8/javafx/interoperability-tutorial/concurrency.htm#BABCHBEA):
> There is no reliable way in Java to stop a thread in process. However, the task must stop processing whenever cancel is called on the task. The task is supposed to check periodically during its work whether it was cancelled by using the `isCancelled` method within the body of the call method.

Along with some minor improvements I've also written a test ( #501 ) proving that task cancellation works. Not sure why would you need to store anything in a `ThreadLocal`?

I don't think that `Service` would be fitting here. I believe that those _services_ are meant to wrap a particular tasks, but in this case more generic approach is necessary. And, well, it's already quite confusing, I'd rather avoid adding another layer of complication.

Romanow88:
@genobis thanks for the test. 

In my tasks I go heavy file system operation in a set of classes (4-5). So my thread does not sleep, it;s usually always busy - which means no interrupted exception. I store the task in a ThreadLocal so I don't need to pass it trough more than 5 method calls to be able to check the task status in points where I do the heavy IO.

edvin:
I don't think that adds much value, you need to check if the task is cancelled obviously, but that's not something you should do in a tight loop. Check it once a second or whatever is appropriate and you should be fine.

PR #499, opened 2017-10-17
Restored runAsync in Component.kt to state prior to PR #496

genobis:
To avoid breaking API

### DIFF
--- a/src/main/java/tornadofx/Component.kt
+++ b/src/main/java/tornadofx/Component.kt
@@ -262,7 +262,10 @@ abstract class Component : Configurable {
         return prop.set(injectable, value)
     }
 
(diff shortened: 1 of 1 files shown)


edvin:
Thank you! And thanks again for this initiative :)

genobis:
@edvin Thank **you**.

Without TFX, working with JFX would be *much* more painful (and it's one of the lightest terms that came into my head). It fells nice to contribute, especially to project managed so well :)

edvin:
That's how I feel as well :) It feels like we're adding real value to JavaFX with this project :)

I'm very happy for your contributions, right now I have limited time to contribute myself, so I'm grateful for everybody that helps with picking up the slack in the mean time!

PR #501, opened 2017-10-20
Fix/async

genobis:
That `dummyUncaughtExceptionHandler` may not look pretty, but at least it lets us avoid a NPE if no exception handler is set (and usually it's not...)

Fixed some ugliness in FXTask - now it's cleaner and safer (`completedProperty` is now truly immutable).

Added a test for task cancelling.

### DIFF
--- a/src/main/java/tornadofx/Async.kt
+++ b/src/main/java/tornadofx/Async.kt
@@ -18,6 +18,11 @@ import java.util.concurrent.Executors
 import java.util.concurrent.ThreadFactory
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicLong
+import java.util.logging.Level
+import java.util.logging.Logger
+
+internal val log = Logger.getLogger("tornadofx.async")
+internal val dummyUncaughtExceptionHandler = Thread.UncaughtExceptionHandler { t, e -> log.log(Level.WARNING, e) { "Exception in ${t?.name ?: "?"}: ${e?.message ?: "?"}" } }
 
 internal val tfxThreadPool = Executors.newCachedThreadPool(object : ThreadFactory {
     private val threadCounter = AtomicLong(0L)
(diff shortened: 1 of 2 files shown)


genobis:
I'll look into it some more. Either the test is flawed (quite possible), or for some reason `cancel` isn't as reliable as interruption (which, for the very similar test case, worked perfectly fine for me).

genobis:
@edvin I removed a problematic test case, but I wonder - am I missing something very obvious, or something is actually wrong somewhere?  I'll admit, I'm not super familiar with JavaFX concurrency, it's built on top on well-known base, but it also adds some new things. Some race/deadlock apparently occurs, and it has *something* to do with cancelling, as if task just stopped running...

Three things to consider:
- It doesn't fail always, just *sometimes*, which unfortunately means often enough to be worried (or it's a proof that this case is flawed)
- The same logic, but based on thread and interruption, works perfectly fine
- Exactly same thing happens on 793b080b68309c8633eac9fc6e2a4adec375eae8 (just checked it), so at least I know I didn't introduce the problem (if it *is* a problem at all).

```kotlin
    @Test
    fun taskCancel() {
        val latch = Latch()
        val holder = object { var v = false }

        val task = runAsync {
            while(!this.isCancelled) {
                try {
                    Thread.sleep(500)
                }
                catch(ie: InterruptedException) {
                    //may happen or not, doesn't matter
                    break
                }
            }
            holder.v = true
            latch.release()
        }

        assertFalse(task.isCancelled)
        task.cancel(true)
        assertTrue(task.isCancelled)
        latch.await()
        assertTrue(holder.v)
    }
```

edvin:
I think it's problematic to check for task.isCancelled with a latch like this. I don't think there is a problem here, just that you can't reliably check for the cancelled status like this. You might have better luck using listeners on the cancelled property though.

PR #506, opened 2017-10-23
Fixes #495 and more tests

### LINKED ISSUE #495: Collections.kt moveUpAt & moveDownAt
I am writing tests for method in Collections.kt and I run into something confusing:

```
/**
 * Moves the given element at specified index up the **MutableList** by one increment
 * unless it is at the top already which will result in no movement
 */
fun <T> MutableList<T>.moveUpAt(index: Int) {
    if (index == 0) return
    if (index !in 0 until size) throw Exception("Invalid index $index for MutableList of size $size")
    val newIndex = index + 1
    val item = this[index]
    removeAt(index)
    add(newIndex, item)
} 
```
From the comment it should move the item UP or do nothing if it is at the top. Assuming `if (index == 0) return` is the check for the top the method is wrong. It is moving the item away from the top (index 0).

For `moveDownAt` the check for bottom is `if (index == size - 1) return` but the item is moved closer to 0.

Is the behavior correct?

There is also a moveUp method:
```
/**
 * Moves the given element **T** up the **MutableList** by an index increment
 * unless it is at the top already which will result in no movement.
 * Returns a `Boolean` indicating if move was successful
 */
fun <T> MutableList<T>.moveUp(item: T): Boolean {
    val currentIndex = indexOf(item)
    if (currentIndex == -1) return false
    val newIndex = (currentIndex - 1)
    if (currentIndex <= 0) return false
    remove(item)
    add(newIndex, item)
    return true
}
```
where up is `val newIndex = (currentIndex - 1)`

Which position is the top of the list 0 or size-1?

LordRaydenMK:
Fixes #495 as discussed.

I activated the CollectionsTest tests (they were commented out).

I also used Preconditions.check for checking the index is valid to throw IllegalStateException (as the other functions).

Added tests for moveAll function.

### DIFF
--- a/src/main/java/tornadofx/Collections.kt
+++ b/src/main/java/tornadofx/Collections.kt
@@ -51,8 +51,8 @@ fun <T> MutableList<T>.moveAll(newIndex: Int, predicate: (T) -> Boolean) {
  */
 fun <T> MutableList<T>.moveUpAt(index: Int) {
     if (index == 0) return
-    if (index !in 0 until size) throw Exception("Invalid index $index for MutableList of size $size")
-    val newIndex = index + 1
+    check(index in indices, { "Invalid index $index for MutableList of size $size" })
+    val newIndex = index - 1
     val item = this[index]
     removeAt(index)
     add(newIndex, item)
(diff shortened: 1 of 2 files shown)


edvin:
Perfect, thanks!

Issue #507, opened 2017-10-23
Tornadofx thread pool behavior


anidotnet:
It would be good if you expose some mechanism to override tornadofx's default thread pool factory. By default all threads are non-daemon threads, where as I want them to be all daemon threads. With current api I can't find a way to achieve it. (The reason behind, my applications does not exit as soon as I close it. It takes around 1 min to close, whereas I have not scheduled in long running job)

edvin:
Good call, this is definitely needed!

edvin:
I've added a `ThreadPoolConfig` object so that you can configure this for example in the `init` block of your app class:

```kotlin
ThreadPoolConfig.daemonThreads = true
```

Does that work for you?

edvin:
I'd also like to investigate the viability of setting this as true by default.

bekwam:
Does System.exit() close your application quicker?

anidotnet:
Can we not add another boolean parameter in ```runAsync()``` method, where true means it will create daemon thread and false will create non-daemon one?

Something like this?

```kotlin
fun <T> task(daemon: Boolean = true, taskStatus: TaskStatus? = null, func: FXTask<*>.() -> T): Task<T> =
        FXTask(taskStatus, func = func).apply {
            setOnFailed({ Thread.getDefaultUncaughtExceptionHandler().uncaughtException(Thread.currentThread(), exception) })
            if (daemon) daemonThreadPool.execute(this)
            else threadPool.execute(this)
        }

fun <T> runAsync(daemon: Boolean = true, status: TaskStatus? = null, func: FXTask<*>.() -> T) = task(daemon, status, func)
```

bekwam:
I think =false might be a better default since you might use a non-daemon Thread to complete a shutdown operation or finish up an operation.  That way, if you call Platform.exit() or close the last window your app won't exit immediately and leave the currently running or ending operations in a bad state.

AlmasB:
+1 for `ThreadPoolConfig.daemonThreads` being `false` by default, as discussed by above comment. 

Regards API allowing `task` and `async` to run as daemon, I'd suggest `taskDaemon` or `daemonTask`, same goes with `async`. The motivation being to lift the boolean parameter off the call, since in calls `runAsync(true, ...)` and `task(false, ...)` it is not immediately clear from the context what the boolean parameter refers to.

edvin:
@AlmasB I don't think that's a big issue in Kotlin since we have named parameters.

[...]

PR #508, opened 2017-10-23
allow using both daemon and non-daemon async tasks, #507

AlmasB:
Double check the logic but as I understand API doesn't break + the behavior remains the same, as previously it was all non-daemon, right?


edvin:
This looks good, but I wonder if it would break API if a user had previously done `runAsync(status) { ... }`. In any case, that's an easy fix, so I think it's worth it. I'll be sure to mention it in the release notes once I've tested some existing calls.

PR #509, opened 2017-10-24
Support for daemon and non-daemon thread in runAsync - #507

anidotnet:


edvin:
Thank you for this :) I merged the other PR but added your async test from this PR :)

PR #510, opened 2017-10-24
Terminate executors in shutdown hook

anidotnet:


edvin:
Perfect! Thank you so much :)

PR #516, opened 2017-10-26
Update Nodes.kt

picoworm:
Hi guys!

I offer to make these functions inlined. So Kotlin will not generate extra classes for lambdas.
This will workaround one kotlin issue with broken links inside bytecode files.
I deal with this issues when trying to Proguard my code that uses anchorPaneConstraints {} and hBoxConstraints {}
Also inlining will give some tiny speedup


edvin:
I don't think this will result in a speedup, but if it solves your proguard issue I don't think it would hurt :) Merging!

PR #551, opened 2017-11-26
Child interceptor OSGI

sleonidy:
Added registration of child Interceptors through osgi.


edvin:
Perfect! Thank you :))

PR #559, opened 2017-12-01
Eliminate interactions between tests

kastork:
On some systems, the order of test execution could impact tests passing or failing, indicating that state from one test could be carried over to other tests.

This commit does two things to try and avoid that situation.

1. Sets up the Maven Surefire plugin to run classes in separate forks (JVMs).

2. Refactor the two problematic tests to use the same approach for setting up and tearing down the FX environment they run in.

On my macOS system all tests now pass when running from Maven and also when running directly in IDEA.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -258,7 +258,21 @@
                     </execution>
                 </executions>
             </plugin>
-        </plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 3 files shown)


kastork:
erg, sorry about the white space!

Maybe we need an editor style file in the project.

edvin:
Thank you!

PR #561, opened 2017-12-06
use kotlin-stdlib-jdk8 instead of jre8

unlimitedsola:
Reference: http://kotlinlang.org/docs/reference/whatsnew12.html#kotlin-standard-library-artifacts-and-split-packages


edvin:
Thanks. We're not supporting Java 9 yet, but we might as well change the package dependency I guess.

Issue #581, opened 2017-12-31
TreeTableView<S>.nestedColumn doesn't work correctly 


DaanVandenBosch:
I think this is because subsequent `TreeTableView<S>.column` calls don't always call `addColumnInternal `which uses the property `"tornadofx.columnTarget" `to know which column to add the nested columns to.

edvin:
Thanks for the PR :)

PR #582, opened 2017-12-31
TreeTableView<S>.nestedColumn fix (#581)

DaanVandenBosch:
TreeTableView<S>.nestedColumn in combination with TreeTableView<S>.column now works correctly for all versions of TreeTableView<S>.column (#581).


edvin:
Thanks for this :)

PR #590, opened 2018-01-14
ItemViewModel.item nullable instead of ambiguous platform type.

dominiks:
At the moment the `item` field of ItemViewModel uses the platform type with uncertain nullability. This can result in NPEs since the user is not aware that this value can be `null` and the compiler does not care about that for platform types.

### DIFF
--- a/src/main/java/tornadofx/Async.kt
+++ b/src/main/java/tornadofx/Async.kt
@@ -455,9 +455,9 @@ class FXTask<T>(val status: TaskStatus? = null, val func: FXTask<*>.() -> T) : T
 }
 
 open class TaskStatus : ItemViewModel<FXTask<*>>() {
-    val running: ReadOnlyBooleanProperty = bind { SimpleBooleanProperty().apply { if (item != null) bind(item.runningProperty()) } }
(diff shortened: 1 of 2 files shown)


edvin:
This would be a breaking change, and the behavior is in line with how property delegates work. I'm reluctant to change this.

dominiks:
It would break all uses of `item` that ignore that the value might be null and are in danger of throwing an NPE at runtime at some point. I feel that leaving it as it is would ignore an important feature of Kotlin.

The behaviour of the property delegates reflects the platform types of JavaFX's properties and until we have a proper non-nullable JavaFX property they should be treated as nullable.

dominiks:
I'll close this then until I find a better way to not have the platform types flying around in my Kotlin code ;-)

kastork:
I'd like to know more about the details of how things would get broken here.

My (possibly naive) assessment is that everywhere someone says

`model.item.something`

The compiler would start complaining and force you to say

`model?.item?.something`

or

`model!!.item.something`

If, and I realize it is a big if, this is the extent of the breakage, then I don't see that breakage as all that catastrophic for users.  It would be a matter of making something that should never have compiled in the first place actually not compile.  I can say from personal experience that saying `model.item.something` in my code has resulted in unwelcome runtime errors that I wish the compiler had told me about.

Is there more to the breakage than this?

PR #607, opened 2018-01-25
Added ResourceBundle.format function to insert formatted strings.

dominiks:
Allows to quickly insert translation strings that contain placeholders as specified in https://docs.oracle.com/javase/tutorial/i18n/format/messageFormat.html.


edvin:
Nice one :)

PR #616, opened 2018-01-30
Add test case for multiple lifecycle async app bug

carltonwhitehead:
@edvin already committed a fix for this issue I had posted about at https://stackoverflow.com/questions/48512919/java-util-concurrent-rejectedexecutionexception-when-action-calls-runasync

I ran this test against a snapshot with his change and confirmed the issue is fixed

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,6 +70,12 @@
             <version>${de.jensd.fontawesomefx.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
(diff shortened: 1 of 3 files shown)


edvin:
Thank you! Can you rename MainView and MainController to something more specific to avoid potential clashes with other tests?

carltonwhitehead:
Yes, will do

edvin:
Perfect!

PR #620, opened 2018-02-01
Remove maven surefire configuration preventing tests from running

carltonwhitehead:
Surefire is now executing tests on my machine. Only one test is failing:
ChildInterceptorTest.

Opening this PR prior to fixing the one failing test to solicit feedback on the `pom.xml` change.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -264,21 +264,13 @@
                     </execution>
                 </executions>
             </plugin>
-			<plugin>
-				<groupId>org.apache.maven.plugins</groupId>
-				<artifactId>maven-surefire-plugin</artifactId>
-				<version>2.20.1</version>
-				<configuration>
-					<!--<parallel>classes</parallel>-->
-					<forkCount>1</forkCount>
-					<reuseForks>false</reuseForks>
-					<includes>
-						<include>**/tests/*Test.kt</include>
-					</includes>
-				</configuration>
-			</plugin>
+            <plugin>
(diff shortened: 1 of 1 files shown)


carltonwhitehead:
Quite surprising to see the Travis CI job didn't fail on ChildInterceptorTest.

I suggest reviewers compare the logs for this PR build,  [the build for the last commit prior to the surefire tests no longer executing](https://travis-ci.org/edvin/tornadofx/builds/308646908?utm_source=github_status&utm_medium=notification), and [the first build where surefire tests began no longer executing](https://travis-ci.org/edvin/tornadofx/builds/310282703?utm_source=github_status&utm_medium=notification). When surefire runs the tests, it logs `Tests run:` for each suite, and again for the sum after all have finished.

kastork:
That failing test was the one that my original PR was meant to address.  I think it is good that this is being looked at since there have been an increasing number of questions on slack about doing gui testing -- it will be good if we can just point to the TornadoFX repo to provide good approaches for that.

It is essential that tests do not interact with each other so that all environments produce the same test results. (different environments appear to run tests in different orders).

carltonwhitehead:
After merging the latest changes, the tests are now running and passing consistently on my Linux laptop.  It would be good to check with some other platforms now. If they're passing too, I think this is ready to merge.

edvin:
Sweet, then we have tested Windows, Mac and Linux. In any case, the order of tests should no longer cause problems. Merging :)

PR #628, opened 2018-02-08
-failFast validation added

tieskedh:


tieskedh:
just discovered it. I like the failing test...

PR #632, opened 2018-02-09
Replaced some `!!` operators with val, ?.let, non-platform types or more meaningful exceptions

r4zzz4k:
There are some places where these weren't needed at all, and in others actual `KotlinNullPointerException` probably may never appear because of the nature of JavaFX single-threaded property access. Nevertheless I've decided to work around some places where this operator is used so that either there are no possibilities for a race condition or user gets some meaningful error message.

Feel free to point me at places which can definitely be left intact :)

### DIFF
--- a/src/main/java/tornadofx/ItemControls.kt
+++ b/src/main/java/tornadofx/ItemControls.kt
@@ -261,14 +261,12 @@ class LazyTreeItem<T : Any>(
         val itemProcessor: (LazyTreeItem<T>) -> Unit = {},
         val childFactory: (TreeItem<T>) -> List<T>?
 ) : TreeItem<T>(value) {
-    var leafResult: Boolean? = null
+    private val leafResult: Boolean by lazy { leafCheck(this) }
     var childFactoryInvoked = false
     var childFactoryResult: List<T>? = null
 
     override fun isLeaf(): Boolean {
-        if (leafResult == null)
-            leafResult = leafCheck(this)
-        return leafResult!!
+        return leafResult
     }
 
     override fun getChildren(): ObservableList<TreeItem<T>> {
@@ -311,8 +309,11 @@ class LazyTreeItem<T : Any>(
     private fun invokeAndSetChildFactorySynchronously(): List<T>? {
         if (!childFactoryInvoked) {
             childFactoryInvoked = true
-            childFactoryResult = childFactory(this)
-            if (childFactoryResult != null) super.getChildren().setAll(childFactoryResult!!.map { newLazyTreeItem(it) })
+            childFactoryResult = childFactory(this).also { result ->
+                if(result != null) {
+                    super.getChildren().setAll(result.map { newLazyTreeItem(it) })
+                }
+            }
         }
         return childFactoryResult
     }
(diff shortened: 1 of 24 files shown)


src/main/java/tornadofx/LayoutDebugger.kt
```diff
@@ -63,7 +63,7 @@ class LayoutDebugger : Fragment() {
 
     private fun hookListeners() {
         nodeTree.selectionModel.selectedItemProperty().onChange {
-            if (it?.value != null) setSelectedNode(it!!.value.node)
+            if (it?.value != null) setSelectedNode(it.value.node)
```
tieskedh: I'm not sure, but maybe use let, like you do on all the other places?

src/main/java/tornadofx/ItemControls.kt
```diff
@@ -261,14 +261,12 @@ class LazyTreeItem<T : Any>(
...
-    var leafResult: Boolean? = null
+    private val leafResult: Boolean by lazy { leafCheck(this) }
     var childFactoryInvoked = false
     var childFactoryResult: List<T>? = null
 
     override fun isLeaf(): Boolean {
-        if (leafResult == null)
-            leafResult = leafCheck(this)
-        return leafResult!!
+        return leafResult
```
tieskedh: remove the body: `override fun isLeaf() = leafResult`

edvin:
Very nice, thanks a lot! Sorry it took me a while to get through it. I'll merge as is, but I think @tieskedh's suggestions are good points, feel free to submit a PR for those remaining changes :)

src/main/java/tornadofx/ItemControls.kt
```diff
@@ -311,8 +309,11 @@ class LazyTreeItem<T : Any>(
     private fun invokeAndSetChildFactorySynchronously(): List<T>? {
         if (!childFactoryInvoked) {
             childFactoryInvoked = true
-            childFactoryResult = childFactory(this)
-            if (childFactoryResult != null) super.getChildren().setAll(childFactoryResult!!.map { newLazyTreeItem(it) })
+            childFactoryResult = childFactory(this).also { result ->
```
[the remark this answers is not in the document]
r4zzz4k: From what I see `childFactory` is not nullable, are you sure we need `?.` here?

src/main/java/tornadofx/LayoutDebugger.kt
r4zzz4k: If `it` is `null` then the condition will fail and we won't get to the `setSelectedNode` call.

src/main/java/tornadofx/ItemControls.kt
r4zzz4k: Sounds reasonable. Should I fill new PR with only this one change?
There are other places where some small refactorings could be done, so maybe this can be left as is right now.

r4zzz4k:
Sorry for not appearing earlier, I should've get to comments before the merge.

There are still places where `!!` is left intact as I was unsure what should be done there -- mostly nullability was guaranteed by other means or there really was nothing to do but fail as the application state doesn't have any sense. Maybe these should be revisited anyway at some point in the future.

Anyway, it's nice to be of use for this great project, even with such a small contribution. Keep up the good work!

edvin:
Yeah, I've found places where `!!` seems to make the most sense, really. Very happy for your contribution, and `!!` often looks like a code smell, so it was nice to get these outta there :)

PR #639, opened 2018-02-14
return if replaced with when.

tieskedh:
gives shorter lines, without losing clear body, like if/else would do


edvin:
That's a lot easier on the eyes :)

PR #643, opened 2018-02-16
adds internal fun reduceSorted for a smarter SmartResize

tieskedh:
This filters only once for it.resizeType.isResizable


tieskedh:
can I use a sortedSet?

PR #661, opened 2018-03-05
Dialog Function that returns an item of type <T>

bkenn:
This new dialog function creates a dialog and returns value . It uses a ChildInterceptor to make the body of the dialog function be the content of the dialog. There are multiple ways to use the dialog. You can totally bypass the setResultConverter and use a ItemViewModel if you create the dialog in the same view. There are two different approach in DialogAppTest.

### DIFF
--- a/src/main/java/tornadofx/Dialogs.kt
+++ b/src/main/java/tornadofx/Dialogs.kt
@@ -1,5 +1,7 @@
 package tornadofx
 
+import javafx.event.EventTarget
+import javafx.scene.Node
 import javafx.scene.control.Alert
 import javafx.scene.control.ButtonType
 import javafx.scene.control.Dialog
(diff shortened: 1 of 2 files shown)


edvin:
We already have a dialog builder so we should first investigate if this can be merged with the existing builder somehow.

bkenn:
You're right. I missed that. I was only looking in Dialog.kt. I did take a look at the dialog builder in Component.kt and I am wondering what if a user doesn't want to use a Form for their dialog and they have 
some custom ui form they would like to use?

edvin:
We could probably use a more general dialog function, if it's problematic having a form as the outer container for the dialog (in most cases it won't be an issue, even if you don't create fieldsets inside of it). Returning a function from a dialog is a bit .. "imperative", and often not the most elegant solution, so you might be better off using `builderWindow` or `builderFragment` for example. I'd like a more general ways for UIComponents to return a value once they are closed though. This is easy to do now, but there should be an easily discoverable way to do it.

PR #668, opened 2018-03-13
Added initialDirectory to chooseFile

razzlero:
This allows specifying an initial directory to look in when using the file chooser.
I added it as the last parameter to avoid breaking code which uses positional arguments.

### DIFF
--- a/src/main/java/tornadofx/Dialogs.kt
+++ b/src/main/java/tornadofx/Dialogs.kt
@@ -70,7 +70,14 @@ enum class FileChooserMode { None, Single, Multi, Save }
  *
  * If the user cancels, the returnedfile list will be empty.
  */
-fun chooseFile(title: String? = null, filters: Array<FileChooser.ExtensionFilter>, mode: FileChooserMode = Single, owner: Window? = null, op: FileChooser.() -> Unit = {}, initialDirectory: File? = null): List<File> {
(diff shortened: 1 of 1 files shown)


edvin:
This both breaks backwards compatibility and makes it impossible to use the op parameter in a practical way. You could instead add the parameter before the op, and make an overloaded function that calls this with null for this parameter, to keep backwards compatibility.

razzlero:
Hi, sorry I'm relatively new to Kotlin. I haven't fully read up how lambdas or higher order functions work in Kotlin, so I was under the impression that it was safe as long as it was added at the end and had a default parameter. 

In terms of overloading I see a slight issue. If we are passing named parameters it could cause an ambiguous overload. I see 2 options to overcome this. 1 is to give the parameters different names in one of the functions. The other option is to give the function with the initialDirectoty parameter a different name. Giving one of the functions a different name is probably the lesser of 2 evils because we would only be renaming 1 thing instead of multiple things. I'll upload a commit implementing it so that you can see what I mean.

razzlero:
Now that I understand better how the op parameter works I realise that it's possible to set the initialDirectory using it with something like this:

```
var fileList = chooseFile(title = "title", filters = filters) {
                            initialDirectory = File(path)
                        }
```

This looks like a better solution, so I'll cancel this pull request.

PR #684, opened 2018-03-27
Css dsl relation

NekoiNemo:
Added ability to use all relations inside CssSelectionBlock's block

### DIFF
--- a/src/main/java/tornadofx/CSS.kt
+++ b/src/main/java/tornadofx/CSS.kt
@@ -954,10 +954,10 @@ class CssSelectionBlock(op: CssSelectionBlock.() -> Unit) : PropertyHolder(), Se
 
     private fun addRelation(
         relation: CssSubRule.Relation,
-        selector: Selectable, selectors: Array<out Selectable>,
+        selector: Selectable, vararg selectors: Selectable,
         op: CssSelectionBlock.() -> Unit
     ): CssSelection {
-        val s = select(selector, *selectors)(op)
+        val s = select(selector, *selectors, op = op)
         selections[s] = relation
         return s
     }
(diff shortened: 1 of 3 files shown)


src/main/java/tornadofx/CSS.kt
```diff
@@ -953,13 +952,69 @@ class CssSelectionBlock(op: CssSelectionBlock.() -> Unit) : PropertyHolder(), Se
     @Deprecated("Use and() instead as it's clearer", ReplaceWith("and(selector, *selectors, op = op)"))
     fun add(selector: Selectable, vararg selectors: Selectable, op: CssSelectionBlock.() -> Unit) = and(selector, *selectors, op = op)
 
-    fun and(selector: String, op: CssSelectionBlock.() -> Unit) = and(selector.toSelector(), op = op)
-    fun and(selector: Selectable, vararg selectors: Selectable, op: CssSelectionBlock.() -> Unit): CssSelection {
+    private fun addRelation(
+        relation: CssSubRule.Relation,
+        selector: Selectable, selectors: Array<out Selectable>,
```
ruckustboom: This should be `selector: Selectable, vararg selectors: Selectable` and you would call this function with `addRelation(<relation>, selector, *selectors, op = op)`

src/main/java/tornadofx/CSS.kt
```diff
@@ -953,13 +952,69 @@ class CssSelectionBlock(op: CssSelectionBlock.() -> Unit) : PropertyHolder(), Se
...
+
+    /**
+     * [CssSubRule.Relation.DESCENDANT]
+     */
+    fun contains(selector: String, op: CssSelectionBlock.() -> Unit) = contains(selector.toSelector(), op = op)
+
+    /**
+     * [CssSubRule.Relation.DESCENDANT]
+     */
+    fun contains(selector: Selectable, vararg selectors: Selectable, op: CssSelectionBlock.() -> Unit): CssSelection =
```
ruckustboom: @edvin Do you think we should add the `contains` function? It is more complete, but it is also the default functionality if you don't use a function.

ruckustboom: Never mind, it's good for constancy.

src/main/java/tornadofx/CSS.kt
NekoiNemo: Is there any reason to use additional spread operator?

ruckustboom: Yes. For one thing, we've been trying to push using idiomatic code throughout the codebase. For another, it makes the intended use clear and ensures consistency,

edvin:
Perfect, thanks a lot!

ruckustboom:
Looks great! I hope we keep seeing more contributions from you :)

PR #685, opened 2018-03-27
added initialDirectory support for chooseFile()

abishai:
initialDirectory support for chooseFile()


edvin:
This would break the API, and it's just as easy to do:

```kotlin
chooseFile(..) {
    initialDirectory = ...
}
```

abishai:
Ah, OK. I'm not very familiar with Kotlin/tornadofx yet, but chooseDirectory() has it.

edvin:
Yeah, I'd love to add it as a parameter to align the two, but unfortunately that would break backwards compatibility, so we'll have to live with this until we do a new major release. I'm thinking about TornadoFX 2.0 when JavaFX has been decoupled from the JDK core.

PR #703, opened 2018-04-16
refactoring bindings

tieskedh:
- replaced some ifs
- placed some breaks
- removed some method-bodies
- ignored some unused parameters
- made some internalBind-references
- removed some unnecessary variables
- introduced some bugs?


edvin:
Thanks a lot, Thijs!

PR #705, opened 2018-04-17
replaced opcr with attachtTo

tieskedh:


tieskedh:
I think the not passing is not related to the code?

edvin:
Yeah, it happens with Travis some times :) Merged. I'll add a comment to the changelog, since I know people have been using `opcr`.

PR #711, opened 2018-04-27
Remove usage of HostServicesFactory

garyttierney:
Fixes builds under OpenJDK.

### DIFF
--- a/src/main/java/tornadofx/Component.kt
+++ b/src/main/java/tornadofx/Component.kt
@@ -2,7 +2,6 @@
 
 package tornadofx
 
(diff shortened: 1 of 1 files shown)


garyttierney:
It could be switched to `HostServices` from the JavaFX package (which is probably preferable), but that'd be a breaking change since `HostServices` and `HostServicesDelegate` unfortunately don't have a compatible API.

edvin:
I believe the new way to do it is via Application.getHostServices(). I’ve commited a fix. Sorry, didn’t check my Mail before committing, so I didn’t see your PR :)

Issue #735, opened 2018-06-07
Canceling Wizard with X doesn't call onCancel()


SackCastellon:
While using the Wizard I have noticed that when canceling using the top-right X button the `onCancel()` function is only called the first time the Wizard is shown, when the new instance is created.

When the wizard is reopened using `find<MyWizard>()`, and the previously created instance is reused, if you press the X button the `onCancel()` funtion is not called.

I suspect the problem is that the `setOnCloseRequest()` in the Wizard's `init` [(Wizard.kt#L223-L226)](https://github.com/edvin/tornadofx/blob/master/src/main/java/tornadofx/Wizard.kt#L223-L226) only affects the stage shown the first time.

### Environment

- **Java:** jdk1.8.0_171
- **TornadoFX:** 1.7.17-SNAPSHOT

### Test code
```kotlin
package wizard

import tornadofx.*

fun main(args: Array<String>) {
    launch<WizardBug>()
}

class WizardBug : App(Main::class)

class Main : View() {
    override val root = borderpane {
        setPrefSize(200.0, 150.0)
        center { button("Wizard").action { find<MyWizard>().openModal() } }
    }

}

class MyWizard : Wizard() {

    init {
        add(WizardPage::class)
    }

    override fun onCancel() {
        println("onCancel")
        super.onCancel()
    }
}

class WizardPage : View() {
    override val root = borderpane { center { label("Hello World") } }
}
```

### Test video
![Gif showing bug](https://user-images.githubusercontent.com/5330355/41128865-ebfa3fbc-6aaf-11e8-8d42-bfd69be8c2f9.gif)

A possible solution could be to re-run the `setOnCloseRequest()` each time a Wizard view is shown.

edvin:
Thank you, I've merged your PR :)

PR #736, opened 2018-06-07
`onCancel()` is now called when reusing Wizard instance (#735)

SackCastellon:
Fix for #735, now each time the scene changes the `OnCloseRequest` is reset, so when pressing the X close button the `onCancel()` function is called even if the Wizard instance is reused.

PS: This is a quick fix I came up with, maybe it can be improved


edvin:
Thanks, merged :)

PR #751, opened 2018-07-10
Simplified the code to convert the params key to string

SackCastellon:


edvin:
I'm not sure this makes sense to include. Can you show me the code you need to write without this function and how it would look with it?

SackCastellon:
Oh, no I don't need this for something to work. This is just that I saw the code and noticed that it could  be simplified by using a high function from the standard library. Both the new and the old code do exactly the same.

SackCastellon:
Also I just noticed I may be possible to improve it a bit more by destructuring the `it` inside the lambda to `(k, _)`

Issue #767, opened 2018-07-31
removeClass() should completely remove the class from styling


ethanmdavidson:
Because JavaFX uses an ObservableList ([Node.styleClass](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/Node.html#getStyleClass--)) to track which style classes are on a node, it is possible to add the same class multiple times. However, the [removeClass extension](https://tornadofx.io/dokka/tornadofx/tornadofx/remove-class.html) in TornadoFX uses [List.remove()](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#remove-java.lang.Object-) which only removes the first occurrence. This means that the following code
```kotlin
button("My Button"){
    addClass(myClass)
    addClass(myClass)
    removeClass(myClass)
}
```
unintuitively results in the button being styled with myClass. I don't have much experience with CSS, but as I understand it [repeated classes on one Node have no effect](https://stackoverflow.com/questions/8828687/css-does-repeating-the-same-class-on-an-element-change-any-behavior). Given this, it seems to make more sense to use removeAll(), like so:
```kotlin
fun <T: Node> T.removeClass(className: String, removeAll: Boolean = true) = apply {
    if (removeAll) styleClass.removeAll(className) else styleClass.remove(classname)
}
```
I think that this is a much more sane default behavior.

edvin:
Sure, even though JavaFX doesn't do this, I agree it makes more sense. Feel free to submit a PR :)

PR #771, opened 2018-08-03
removeClass() now removes all instances of the class by default

ethanmdavidson:
As discussed in [issue 767](https://github.com/edvin/tornadofx/issues/767)

I also changed TableColumnBase.toggleClass() to use the helper methods (same as Node.toggleClass()) instead of directly manipulating styleClass.

This is my first pr so let me know if I did anything wrong.


edvin:
Perfect :)

Issue #785, opened 2018-08-24
togglebutton Builder Needs ObservableValue<String> Version


bekwam:
You can bind to a button's textProperty by passing in an observable.  

togglebutton only accepts a String.

SiebelsTim:
@edvin This can be closed I guess.

edvin:
Done :)

PR #786, opened 2018-08-24
[#785] Added new method for togglebutton ObservableValue<String> param

bekwam:
Created new function for togglebutton{} for ObservableValue param


edvin:
Sweet! For 2.0 we should look at a way to merge or combine these duplicates.

PR #791, opened 2018-08-27
TornadoFX JDK9+ compatibility

abhinayagarwal:
This PR removes `module-info.java` and adds JavaFX maven artifacts as dependency. Also updates other dependencies and plugin versions.

Since Kotlin's JVM target is set to 1.8, I have set TornadoFXs compiler source and target set to 9 instead of 11, to enable it to be compiled and installed against JDK 9/10/11. 

However, since JavaFX dependencies are compiled against OpenJDK 10, they will fail if someone uses TorandoFX against OpenJDK 9. But, it will work with OracleJDK 9 since bundled JavaFX is used instead of the dependencies.

Here is a compatibility table for projects where TornadoFX is added an dependency:

<table>
<tr>
<td></td>
<td>9</td>
<td>10</td>
<td>11</td>
</tr>
<tr>
<td>OracleJDK</td>
<td>&#10004;</td>
<td>&#10004;</td>
<td>&#10004;</td>
</tr>
<tr>
<td>OpenJDK</td>
<td>&#10006;</td>
<td>&#10004;</td>
<td>&#10004;</td>
</tr>
</table>

All my tests were done against [TornadoFX-Samples](https://github.com/edvin/tornadofx-samples). There is a chance that there are still corner cases which aren't covered. Therefore, testing the PR with various projects is essential.


edvin:
There are numerous package renames and changes to public API between JavaFX 8 and 9, so this won't be enough. Supporting JDK9+ is a much larger task, but it's definitely in the pipe line. I think it might send the wrong signal to make it run on JDK9, only to fail at runtime.

abhinayagarwal:
I didn't look at each of the class. A general search for `com.sun` on .kt files didn't return me any results so I thought they were fixed by someone before me. Since the code compiles under JavaFX9, I am guessing most of it was commented out. I will have a look at the code later today. 

If you have a list of controls/files which fail under JDK9, please let me know. It would speed up the whole process :)

abhinayagarwal:
Closing this PR as jdk9 branch is too old and results in a lot of conflict when merged with master. It would be easier for me to create a new branch from master and work on it. I will create an issue to discuss JDK10/11 support.

edvin:
Good call.

PR #797, opened 2018-09-03
JDK 10/11 compatibility changes

### LINKED ISSUE #794: JDK 10 and beyond
Running master on JDK 10. Here are some issues which I am not sure how to proceed with or need feedback on my current approach. Therefore, I am writing them here to discuss:

**1. ApplicationListener.kt**

Uses reflection to fetch on `com.sun.javafx.application.PlatformImpl`'s `initialized` variable. 

```
val initializedField = PlatformImpl::class.java.getDeclaredField("initialized")
```
My inexperience with osgi tells me we can replace `PlatformImpl` with:

```
Class.forName("com.sun.javafx.application.PlatformImpl").getDeclaredField("initialized")
```

But I do not like this solution. @edvin Does [`Platform#startup`](https://docs.oracle.com/javase/10/docs/api/javafx/application/Platform.html#startup(java.lang.Runnable)) help somehow?

**2. AutoCompleteComboBoxSkin.kt**

`init { }` sets an event handler for mouse click event on `arrowButton`. This node was previously `protected`.

```
arrowButton.setOnMouseClicked {
    if (isShowing) {
        resetFilter()
    }
}
```

I am hoping we could get rid of this code without any implications?

**3. ItemControls.kt**

`DirtyDecoratingTableRowSkin` and `ExpandableTableRowSkin` cannot extend from `TableRowSkin` because of a Kotlin Bug: https://youtrack.jetbrains.com/issue/KT-12255

Our best approach here is to create public Java class for these and let TornadoFX use them. I am open to any other alternative.

**4. SqueezeBox.kt**

`SqueezeBoxSkin` has the following code in `init { }`. The overridden `handleControlPropertyChanged` simply calls `super.handleControlPropertyChanged` (BehaviorSkinBase) which is actually a no-op.

```
protected void handleControlPropertyChanged(String propertyReference) {
    // no-op
}
```
So, may be we can simply remove the init block?

**5. Reflection**

For any other API which has been either changed to private or package-private, I am going to fallback to reflection. This can be discussed in the PR.

**6. Separate branch for JDK10 and beyond**

I would have been much happier if we could keep the same master for developing both JDK8 and JDK10+. This would have been possible if there was no inheritance. Since most uses of skin has a package change, I don't think we could somehow make this possible. If someone has a better plan, now is the time to comment :)

abhinayagarwal:
This PR fixes #794

* For reflection used in `AutoCompleteComboBoxSkin.kt` and `DataGrid.kt`, an application depending on TornadoFX needs to pass `--add-opens javafx.controls/javafx.scene.control.skin=ALL-UNNAMED` when running the application.
* I am using a Java class for `ReflectionUtils` because I was having issues changing the varargs parameter into Kotlin. If someone can help me with it, I would be glad to update it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Fixed
 
+- TableView.onEditStart() and TableColumn.cancel() functions which can be used to intercept editing events
+
 ### Changes
 
 ### Additions
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,13 @@
 
 ### Fixed
 
-- TableView.onEditStart() and TableColumn.cancel() functions which can be used to intercept editing events
-
 ### Changes
 
 ### Additions
+
+- TableView.onEditStart() and TableColumn.cancel() functions which can be used to intercept editing events
+- resources.media() to load a Media instance from resources
+- Media.play() shortcut which creates a MediaPlayer and plays the Media
 
 ## [1.7.17]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- TreeView.`lazypopulate` leafcheck logic was reversed (https://github.com/edvin/tornadofx/issues/773)
 - `onCancel()` is now called when reusing Wizard instance 
 - Observable collection delegates removed because they shadow observable properties (See Properties.kt:L212)
 - The `style` property of `ListCell` will be cleared by the framework so it can be manipulated in `cellFormat` without side effects (https://stackoverflow.com/questions/51459371/custom-cell-format-listview-tornadofx-on-delete-item)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Fixed
 
+- `config` no longer accepts null values, as the underlying `Properties` store won't allow them (https://github.com/edvin/tornadofx/issues/792)
+
 ### Changes
 
 - Kotlin 1.2.61
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Observable collection delegates removed because they shadow observable properties (See Properties.kt:L212)
 - The `style` property of `ListCell` will be cleared by the framework so it can be manipulated in `cellFormat` without side effects (https://stackoverflow.com/questions/51459371/custom-cell-format-listview-tornadofx-on-delete-item)
 - Escape closes window only works first time for Views (https://github.com/edvin/tornadofx/issues/764)
+- Reused modal didn't get stage icons set (https://github.com/edvin/tornadofx/issues/779)
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 - `Window.aboutToBeShown` property avoid false positives for invisible Workspace warning (https://github.com/edvin/tornadofx/issues/755)
 - Slideshow slides supports optional timeout, which will advance to the next slide using the Slide transition
 - `importStylesheet` now supports `file`, `http` and `https` in addition to classpath resources (https://github.com/edvin/tornadofx/issues/762)
+- `raduimenuiutem` builders now accepts `value` property (https://github.com/edvin/tornadofx/issues/737)
 
 ## [1.7.16]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- OnDock Called Twice In openInternalWindow (https://github.com/edvin/tornadofx/issues/772)
 - TreeView.`lazypopulate` leafcheck logic was reversed (https://github.com/edvin/tornadofx/issues/773)
 - `onCancel()` is now called when reusing Wizard instance 
 - Observable collection delegates removed because they shadow observable properties (See Properties.kt:L212)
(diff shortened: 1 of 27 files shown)


abhinayagarwal:
There are tests failing with `java.net.MalformedURLException: unknown protocol: css`.  It might have something to do with TestFX still using JDK 8 version.

```
WARNING: Resource "css://tornadofx.testapps.Styles" not found.
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.164 s <<< FAILURE! - in tornadofx.tests.StylesheetErrorTest
[ERROR] shouldStartApplicationWithWrongStylesheetWithoutCrashing(tornadofx.tests.StylesheetErrorTest)  Time elapsed: 0.014 s  <<< ERROR!
java.lang.NoSuchMethodError: javafx.stage.Window.impl_getWindows()Ljava/util/Iterator;
        at tornadofx.tests.StylesheetErrorTest.shouldStartApplicationWithWrongStylesheetWithoutCrashing(StylesheetErrorTest.kt:41)

[INFO] Running tornadofx.tests.InternalWindowClosingTest
--- Exception in Async Thread ---
java.net.MalformedURLException: unknown protocol: css
```

edvin:
I don't even know if the protocol handler will be supported in JDK10. Take a look at `sun.net.www.protocol.css.Handler` (in the TornadoFX sources). This is supposed to be picked up automatically, but maybe they changed or removed this in JDK10? Needs investigation. We can "save this for last" and disable the test for now IMO.

edvin:
PS: We can't merge this with the master branch as it would break JDK8 obviously. Should we create a new branch for this?

abhinayagarwal:
Yup, creating a new branch sounds good :+1:

abhinayagarwal:
@edvin Is there a reason why TorandoFX has both `maven-javadoc-plugin` and `dokka-maven-plugin`. Can `maven-javadoc-plugin` be used to generate Javadoc for Kotlin?

edvin:
Yes, we need javadoc to pass Maven Central sync requirements :) I don't think `maven-javadoc-plugin` can generate Javadoc for Kotlin.

abhinayagarwal:
@edvin The PR looks at a good state to be merged, unless you want to a look at why `InternalWindowClosingTest` is failing ;)

Do you want me to update the documentation to include that it is necessary open package `javafx.scene.control.skin` using`--add-opens javafx.controls/javafx.scene.control.skin=module-name` to use `DataGrid` and `AutoCompleteComboBoxSkin`?

edvin:
Sweet, let's merge! And yeah, that would be great if you added that to the docs.

abhinayagarwal:
I have replaced the usage of profile with compile time dependencies. OpenJFX dependencies in profile were not resolved as transitive dependencies in a third party project where Tornadofx was added as a dependency.

Doing this has also enabled TorandoFX to run on OpenJDK10/11 and OracleJDK10.

For more information:

* https://issues.apache.org/jira/browse/MNG-1388
* http://maven.40175.n5.nabble.com/Java-11-and-java-xml-bin-etc-td5945156.html#a5945657

abhinayagarwal:
@edvin Can you direct me to the location of the document where the information on `--add-opens` should be added?

edvin:
That's great news! I think we should add it to the README for now. Merge at will :) I'll be able to invest more time into this from some time next week by the way :)

abhinayagarwal:
@edvin I have delaying the merge of this PR because I wanted to create a new travis pipeline for the same. I have been very busy lately ;)

Since Java 8 is nearing EOL, I have been wondering if it even make sense to target this change to a new branch?

abhinayagarwal:
What I meant was that this change is actually targeting to a new branch (jdk10) and may be we should target it to master instead?

edvin:
Ah I see. If it doesn't break Java 8 compatibility we could do that. I was under the impression that it did. Haven't had time to review it myself :)

abhinayagarwal:
It does break Java 8 compatibility.

The reason I raised it is because 8 is getting end of life in [January, 2019](https://www.oracle.com/technetwork/java/javase/eol-135779.html). Therefore, I was wondering if we really want to keep 8 as master.

edvin:
Personal Users will get JDK8 updates until December 2020, but once we have a functional JDK11 version we should swap that for master and move the current master to a jdk8 branch.

abhinayagarwal:
Ok cool 👍 

I hope you get some time real soon to be able to test the PR and approve. In the meantime, I will try to fix the conflicts and setup a TravisCI for 11.

edvin:
Sweet. I feel bad about the lack of time investment lately, but I've had to see this injection molding machine thing through. We made our first plastic part on it yesterday so we're getting there!

altavir:
I would like to vote in favor of moving forward to JDK 9+. For now tornadofx is a real stopper, which prevents me (and probably others) from using modern Java distribution.

abhinayagarwal:
@edvin can you update jdk10 branch with the latest changes from master?
I think the latest changes that I just merged from master branch are causing the conflicts.

thomasnield:
Okay... so are we targeting Java 11 solely? Or 9/10 as well? Trying to figure this out from the POM file.

edvin:
Since Java 9 is not an LTS release, it's already EOL so there is no need to target it. However, if people made the switch to Java 9,  they won't have any issues upgrading to 11 or even 12 :)

edvin:
Fantastic work! Let's get this thing done :) I'll start working on this tomorrow.

abhinayagarwal:
@edvin A gentle reminder - Before making a release, we should test it against different TornadoFX samples so that we don't miss any edge cases :)

edvin:
Absolutely, though there will be breaking changes, so most apps will require some changes I believe. Even if we have to chuck a few features I'd like to get this out there now, but also remove deprecated functions/features and clean up some of the API's. The API should be considered unstable for a little while, and not used in production immediately.

edvin:
Should we go for TornadoFX 2.0 btw? That's what I had in mind, but I'm open for suggestions.

abhinayagarwal:
2.0-RC sounds like a good idea to me ;)

edvin:
Perfect!

abhinayagarwal:
May be we should throw that question in the slack channel for feedback. People are more active in the channel than in this PR :)

altavir:
I think that people are waiting for that. For me personally, the incompatibility with java 9+ is a great stopper.

Issue #801, opened 2018-09-15
Add support to correctly remove the row from the GridPane, as well as remove all rows


zshamrock:
The problem with the removing manually is that you have to use (and so be aware) of the internal `TornadoFX.GridPaneRowId` property which manages the row index internally.

PR #802, opened 2018-09-15
#801 Add support to remove row from the GridPane

zshamrock:
As well as remove all rows.

Currently I extend the `GridPane` manually in my own project https://github.com/zshamrock/dynoman/blob/master/src/main/kotlin/com/akazlou/dynoman/ext/Ext.kt, but would be nice to have these features as part of the TornadoFX package.

Please, let me know whether the provided PR looks good or make any necessary adjustments you feel like good doing.


src/main/java/tornadofx/Layouts.kt
```diff
@@ -33,6 +33,43 @@ fun GridPane.row(title: String? = null, op: Pane.() -> Unit = {}) {
...
+/**
+ * Removes the corresponding row to which this [node] belongs to.
+ *
+ * It does the opposite of the [GridPane.row] cleaning all internal state properly.
+ *
+ * @return the row index of the removed row.
+ */
+fun GridPane.removeRow(node: Node): Int {
+    if (properties.containsKey(GridPaneRowIdKey)) {
+        properties[GridPaneRowIdKey] = properties[GridPaneRowIdKey] as Int - 1
```
zshamrock: Here the alternative implementation could be if the property value is 0, just to remove the property all together instead of subtract.

Although currently it doesn't matter, as in the current case it will result in `-1`, and when adding the new row the index will become `0`, same as if the property would absent.

zshamrock:
Hi, @edvin. Thank you! I am not planning to push any more changes. So if it looks good to you, feel free to proceed with the merge.

edvin:
Great! I will look at it in more detail ASAP and perform the merge! Thanks again :)

edvin:
Thanks again. Merged :)

PR #813, opened 2018-10-01
Add wipe transition

ruckustboom:
Add a wipe that shows how to use a `Canvas` for transitions that are not possible by directly manipulating the nodes.


edvin:
Nice! I'll add it to the CHANGELOG as well :)

ruckustboom:
Thanks, I keep forgetting that

PR #830, opened 2018-11-02
Take advantage of some new functionality and features from Kotlin 1.3

SackCastellon:
This PR updates some code to take advantage of some new Koltin 1.3 features like:
- Contracts`, allowing for proper smart-cast when using:
  - `isNullOrEmpty()`
  - `isNullOrBlank()`
  - `check(Boolean)`
  - `require(Boolean)`
- `hashCode()` with nullable receiver
- Capturing `when` subject in a variable

### DIFF
--- a/src/main/java/tornadofx/LayoutDebugger.kt
+++ b/src/main/java/tornadofx/LayoutDebugger.kt
@@ -310,7 +310,7 @@ class LayoutDebugger : Fragment() {
 
             if (node is Region) {
                 // Background/fills is immutable, so a new background object must be created with the augmented fill
-                if (node.background?.fills?.isNotEmpty() ?: false) {
+                if (!node.background?.fills.isNullOrEmpty()) {
                     field("Background fill") {
                         vbox(3.0) {
                             node.background.fills.forEachIndexed { i, backgroundFill ->
@@ -406,7 +406,7 @@ class LayoutDebugger : Fragment() {
         override fun toString(v: Insets?) = if (v == null) "" else "${v.top.s()} ${v.right.s()} ${v.bottom.s()} ${v.left.s()}"
         override fun fromString(s: String?): Insets {
             try {
-                if (s == null || s.isEmpty()) return Insets.EMPTY
+                if (s.isNullOrBlank()) return Insets.EMPTY
                 val parts = s.split(" ")
                 if (parts.size == 1) return Insets(s.toDouble())
                 if (parts.size == 2) return Insets(parts[0].toDouble(), parts[1].toDouble(), parts[0].toDouble(), parts[1].toDouble())
--- a/src/main/java/tornadofx/Rest.kt
+++ b/src/main/java/tornadofx/Rest.kt
@@ -166,12 +166,10 @@ open class Rest : Controller() {
             try {
                 val content = text()
 
-                if (content == null || content.isEmpty())
+                if (content.isNullOrBlank())
                     return Json.createArrayBuilder().build()
 
-                val json = Json.createReader(StringReader(content)).use { it.read() }
-
-                return when (json) {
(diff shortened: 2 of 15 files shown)


src/main/java/tornadofx/LayoutDebugger.kt
```diff
@@ -406,7 +406,7 @@ class LayoutDebugger : Fragment() {
         override fun toString(v: Insets?) = if (v == null) "" else "${v.top.s()} ${v.right.s()} ${v.bottom.s()} ${v.left.s()}"
         override fun fromString(s: String?): Insets {
             try {
-                if (s == null || s.isEmpty()) return Insets.EMPTY
+                if (s.isNullOrBlank()) return Insets.EMPTY
```
ruckustboom: This is a semantic change. Did you mean `s.isNullOrEmpty()`?

src/main/java/tornadofx/Rest.kt
```diff
@@ -166,12 +166,10 @@ open class Rest : Controller() {
             try {
                 val content = text()
 
-                if (content == null || content.isEmpty())
+                if (content.isNullOrBlank())
```
ruckustboom: This is a semantic change. Did you mean `isNullOrEmpty()`?

src/main/java/tornadofx/LayoutDebugger.kt
ruckustboom: This is a semantic change. Did you mean `isNullOrEmpty()`?

src/main/java/tornadofx/Rest.kt
ruckustboom: This is a semantic change. Did you mean `isNullOrEmpty()`?

src/main/java/tornadofx/Rest.kt
```diff
@@ -187,12 +185,10 @@ open class Rest : Controller() {
...
```
ruckustboom: This is a semantic change. Did you mean `isNullOrEmpty()`?

ruckustboom:
I tried to comment in the code, but GitHub is glitching for me pretty badly, so I'm not sure what you are seeing.

There are a number of places in this PR where `thing == null || thing.isEmpty()` was replaced with `this.isNullOrBlank()`, which is a semantic change.

src/main/java/tornadofx/LayoutDebugger.kt
SackCastellon: You are right it is a semantic change, but in this case, whether the string is empty or blank the result will be `Insets.EMPTY`, in the old code if the string was blank an exception was thrown and then ignored to return `Insets.EMPTY`. In the new code we prevent the exception to be thrown as we know that if it is black we will return `Insets.EMPTY`.

In my opinion this change is justified.

src/main/java/tornadofx/Rest.kt
SackCastellon: In this case, you are right, I think the change is not justified. I will change it to `isNullOrEmpty()`

src/main/java/tornadofx/Rest.kt
SackCastellon: In this case, you are right, I think the change is not justified. I will change it to `isNullOrEmpty()`

edvin:
Thanks guys, good work! Nice to start utilizing the new features of Kotlin 1.3 :)

ruckustboom:
@SackCastellon I agree `isNullOrBlank()` would make more sense in a lot of those places, it's just that it is a semantic change out of scope of the rest of this PR ("Take advantage of some new functionality and features from Kotlin 1.3"). I would recommend creating a second PR with those changes to separate logical corrections from syntactic updates.

PR #836, opened 2018-11-07
Don't start a new thread for each runLater(duration)

mslenc:
In the current implementation, each call to runLater(duration) creates a new Timer, which in turn creates a new Thread. The threads never seem to get cleaned up, and eventually the process runs out of them.

### DIFF
--- a/src/main/java/tornadofx/Async.kt
+++ b/src/main/java/tornadofx/Async.kt
@@ -158,6 +158,8 @@ infix fun <T> Task<T>.cancel(func: () -> Unit) = apply {
  */
 fun runLater(op: () -> Unit) = Platform.runLater(op)
 
+private val timer: Timer by lazy { Timer(true) }
+
 /**
  * Run the specified Runnable on the JavaFX Application Thread after a
  * specified delay.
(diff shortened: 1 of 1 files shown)


mslenc:
Actually, looking at Timer source, it should eventually clean up after itself, but relies on finalize() to do so. I guess the error I saw was the result of my app doing no/very little memory allocation, while calling runLater every 0.2 seconds..
In any case, it's far more efficient to just use a single Timer instance..

edvin:
I agree if you use this a lot it would be better to create the timer once, though if you need to call something like this regularly you should probably not dispatch new jobs all the time, but rather write your own loop around it. If you could rename the `timer` variable to something more specific I will merge :)

mslenc:
Well, it's your name, I just moved the var - https://github.com/edvin/tornadofx/commit/4f5008edc20b04989ee2b4915b1546761aa91277 :)

Anyway, any suggestion on what the name should be? `runLaterTimer`?

edvin:
There is a huge difference between a local variable in a function, and a package level variable, even if it's private :) Yeah, that would be fine.

edvin:
Perfect, thank you!

mslenc:
sorry, I missed the other reference (`FXTimerTask(op, timer)` should be `FXTimerTask(op, runLaterTimer)`)

mslenc:
should I open another pull request?

edvin:
Nah, I'm fixing it and committing now :)

mslenc:
OK, thanks :)

PR #841, opened 2018-11-09
Added Russian translations for ViewModel and Wizard

ValV:
Tiny change for the first pull request (only 2 files added)


edvin:
Nice, thank you :)

PR #875, opened 2018-12-31
Update Workspace.kt

Warkst:
1. The accelerator Ctrl+N is calling `onSave`, should be `onCreate`
2. Updated the accelerators so the keybinds are mac friendly


edvin:
Great! But it seems checks are failing, perhaps a missing import?

Warkst:
Should be fixed :)

PR #890, opened 2019-01-14
Tabpane.creatable bug fix

Warkst:
Found another copy paste bug, Edvin!


edvin:
Ah, those are the best :)

Warkst:
Any plans on when these changes will be available on maven?

edvin:
There is not enough changes to warrant a release right now, but you can use the snapshots from oss.sonatype.org :)

Warkst:
Cool, thanks!

PR #892, opened 2019-01-15
Update to cellFormat

Warkst:
Rather than an additional callback, it might be better to simply reset
the cell's `style` and `styleClass` properties to their defaults.

Now that we're on this topic anyway, I couldn't find `cellFormat` for a
regular `TableColumn` but I did find `SmartTableCell` in `TableView`,
which has a similar style reset in it's `updateItem`, except resetting
the `styleClass` list. Should I create a PR for that, too?


edvin:
Yes please :)

PR #910, opened 2019-01-29
WIP: Some design ideas for adding Guice/Spring support so far

ahinchman1:
Throwing some ideas around. Still need to try this with Dagger as well, which may include a bit more work since I've never tried it with TornadoFX/JavaFX.


ctadlock:
I don’t see the benefit of this given the additional bloat of the new dependencies. Closing. Feel free to reopen with additional justification

PR #926, opened 2019-02-26
New and old value can be null in ChangeListener

greboid:
This probably counts as a breaking change, but it saves getting caught out when you forget.

Would be nice to backport as well, but not sure if you want to change the API in the stable.

### DIFF
--- a/src/main/java/tornadofx/Lib.kt
+++ b/src/main/java/tornadofx/Lib.kt
@@ -221,7 +221,7 @@ fun Clipboard.putString(value: String) = setContent { putString(value) }
 fun Clipboard.putFiles(files: MutableList<File>) = setContent { putFiles(files) }
 fun Clipboard.put(dataFormat: DataFormat, value: Any) = setContent { put(dataFormat, value) }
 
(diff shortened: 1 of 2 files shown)


src/main/java/tornadofx/Lib.kt
```diff
@@ -221,7 +221,7 @@ fun Clipboard.putString(value: String) = setContent { putString(value) }
 fun Clipboard.putFiles(files: MutableList<File>) = setContent { putFiles(files) }
 fun Clipboard.put(dataFormat: DataFormat, value: Any) = setContent { put(dataFormat, value) }
 
-inline fun <T> ChangeListener(crossinline listener: (observable: ObservableValue<out T>?, oldValue: T, newValue: T) -> Unit): ChangeListener<T> =
+inline fun <T> ChangeListener(crossinline listener: (observable: ObservableValue<out T>?, oldValue: T?, newValue: T?) -> Unit): ChangeListener<T> =
```
pavlus: `<T>` has default upper bound and Kotlin generics default upper bound is nullable `Any?`, isn't it?

greboid: But the type comes from Java which doesn't check null, so either value from Java can be null in various listeners, it seems safer to make this explicit rather than implicit

PR #935, opened 2019-03-05
Added convenience function builders for SimpleXXXProperty classes

LittleLightCz:


edvin:
Very nice! Will you add a note to CHANGELOG as well?

edvin:
Perhaps it would make sense to have an overload for listProperty which accepts varargs as well?

edvin:
I'll do it :)

LittleLightCz:
Ahh, sorry for the Changelog :grin: , but thank you :relaxed:

edvin:
No problem, happy for every contribution :)

LittleLightCz:
I am testing the latest snapshot version and there is one thing that hit my eye. For example now when I use `stringProperty("")`, then the inferred type is `SimpleStringProperty`. Therefore if I pass this val to some function, IDEA auto-generates parameter with `SimpleStringProperty` type as well, which pollutes the type name with the "Simple" word again :-)). So I was thinking, is it a good idea for the xxxProperty methods we've just added to return their parent class type instead? Could it cause any issues for some use-cases?

edvin:
It's most definitely better if they return the parent class :)

LittleLightCz:
Ok thanks for advice :-). I will do another PR and will try to fix it.

edvin:
Great, thank you!

PR #949, opened 2019-03-14
Make attachTo() public

tingyik90:
Any consideration to hide this method, since other attachTo() methods are exposed? I'm using this to create custom views.


edvin:
Sure, I've seen several use cases where access to these functions can be beneficial :)

PR #963, opened 2019-03-29
introduces a textfield for kotlin.Int

ronsmits:
and because of that also for java.lang.Integer. This makes it work for Pojo's that are used with "observable"

### DIFF
--- a/src/main/java/tornadofx/Controls.kt
+++ b/src/main/java/tornadofx/Controls.kt
@@ -107,6 +107,11 @@ fun EventTarget.textfield(property: ObservableValue<Number>, op: TextField.() ->
     op(this)
 }
 
+@JvmName("textfieldInt")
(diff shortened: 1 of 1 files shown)


edvin:
What's wrong with `textfield(numberProperty, IntegerStringConverter())`? :)

ronsmits:
There is nothing wrong with it, however (apart from the fact this might need to be documented) it means that a normal pattern of `textfield(property)` needs to be added too. I find that having the setup of an `ItemViewModel` that handles these things make more sense to me, as all the plumping is handled in a single spot (the setting up of the model). The view stays blissfully unaware of these intricaties

edvin:
Thanks, I actually never considered that :) We might extend the Int version to define a default IntegerStringConverter as well perhaps?

ronsmits:
I am currently working on an adaption of the BeanPathAdapter from jfxExtras. Maybe part of the pojo problems can be sorted with that

Issue #969, opened 2019-04-19
Allow to specify the index for the new tab created inside the TabPane


zshamrock:
Currently `TabPane.tab` extensions add the new tab by default to the end of the tabs list. 

Could be good to add a few more extensions where the index can be passed specifying the location of the new tab.

One of the possible use case where it might be handy is when you duplicate the current tab, and want it to be inserted just after the current one, like here https://github.com/zshamrock/dynoman/commit/ef0f45d87435a50d9cad3fdc31d5a5c7edc453a1#diff-8249fde5cb547e044e20aa28212c3ddfR56.

edvin:
If we can add an index parameter without breaking backwards compatibility that would be nice. If not, it's pretty straight forward to do this explicitly:

```kotlin
class MyView : View("My View") {
    override val root = tabpane {
        tab("First") {
            label("I'm first")
        }
        tab("Second") {
            label("I'm second")
        }
        tab("1.5") {
            label("I'm in between")
            tabs.move(this, 1)
        }
    }
}
```

zshamrock:
Will it break the backward compatibility, as I imagined these will be the new separate extensions in addition to the existing ones.

Also thanks for the `move`, didn't notice it there. Use it now instead of `replace & add` (which is actually what `move` does, but just out of the box).

edvin:
Merged :)

PR #971, opened 2019-04-22
#969 Add extra TabPane extension to allow specify the index of the cr…

zshamrock:
…eated tab in the tabs list


zshamrock:
I am not sure how/where the test apps are used, but I also extended the one related to the `TabPane` usage.

Again, feel free to reject the pull request, if you think the above behavior can be achieved by just using `move` method directly.

edvin:
Nice work!

PR #984, opened 2019-05-06
Vector math extensions

TheJhonny007:


edvin:
Kotlin functions should not start with an uppercase letter. Also, several of these would probably be smoother if they were extension functions.

TheJhonny007:
> Kotlin functions should not start with an uppercase letter

The uppercase functions should behave as constructor extensions. IntelliJs autocomplete even suggests them as possible constructors. If you type out `Point2D()` and press **ctrl + p** in the parenthesis IntelliJ suggests the original constructor arguments and the fake constructor arguments.

> Also, several of these would probably be smoother if they were extension functions.

Do you mean the `distance`, `midPoint` and `angle` functions?

edvin:
Aha, I see. In that case I agree. Didn't think of that. The rest actually looks good, I was looking at it on my phone in a truncated state. Sorry about that. Merging :) Thanks a lot!

PR #1008, opened 2019-05-31
Deployment using Travis CI

abhinayagarwal:
This PR adds functionality:

* Deploy snapshots to nexus
* Deploy release when tags are pushed and snapshot is removed from version

Once Travis is enabled, the maintainer needs to add the following environment variables:

• NEXUS_USERNAME
• NEXUS_PASSWORD

### DIFF
--- a/.travis.settings.xml
+++ b/.travis.settings.xml
@@ -1,0 +1,13 @@
+<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0
+                      http://maven.apache.org/xsd/settings-1.0.0.xsd">
+  <servers>
+    <server>
+      <id>sonatype-nexus</id>
+      <username>${NEXUS_USERNAME}</username>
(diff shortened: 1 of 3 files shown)


edvin:
Thank you! I need a little time to verify things before merging.

edvin:
Looks good, now I just need to figure out how to add private environment variables to the build environment :)

edvin:
That was easy :) Very nice!

Setting environment variables from repository settings
$ export NEXUS_USERNAME=[secure]
$ export NEXUS_PASSWORD=[secure]

edvin:
OK, it sort of worked. We also need to export variables for gpg signing, and changing making sure the signing step happens before deployment. I've manually changed sign-artifacts step from deploy to install before publishing releases. I've let it stay as `deploy` in the deployment descriptor so others building it locally wouldn't be required to sign. Any idea how we should handle this?

abhinayagarwal:
For the signing part, we will have to either encrypt the `secret.gpg` and upload it to the repository. Or, we can save encode using BASE64 and save it as env variable. In [OpenJFX maven plugin](https://github.com/openjfx/javafx-maven-plugin/blob/master/.travis.yml), we use the former approach.

We also need to separate the deployment for snapshots and releases depending on what we want to achieve for each step.

Issue #1009, opened 2019-06-05
EventBus leaks subscriptions and scopes


carltonwhitehead:
I noticed in my app's automated tests that subsequent test runs were printing exceptions coming from the `EventBus.fire(event: FXEvent)` exception handler. Nevertheless, my tests were passing, and the exception message and stack trace didn't make any sense with respect to the state I could clearly see my tests were setting up correctly.

With the debugger, I could see the `FX.eventBus.subscriptions` and `FX.eventBus.eventScopes` were growing with the addition of each `@Test` in the suite under which the `EventBus` was in use. Since the event bus is effectively global and `App.stop()` doesn't clear the subscriptions, old subscriptions were leaking beyond the lifecycle of the app, which is torn down after each test. Those old subscriptions' scopes had been deregistered when their app stopped, so when they fire, all the subsequent `find()` and `by inject()` triggered by the old actions getting called was resulting in the nonsense stack traces I was seeing.

As an experiment, locally, I added a `EventBus.clear()` which clears both the `subscriptions` and `eventScopes`, and called it from `App.stop()`. This took care of the issue for my project.

Although I have a code change that fixes the problem, I wanted to open this as an issue first for discussion to rule out any unwanted side effects before I send a PR.

It seems to me that it might also be reasonable to add a method for deregistering all the registrations with a given scope, and to call that from `Scope.deregister()`, in addition to clearing them all when the app stops.

andrey-vasilyev:
Hi

There seems to be a problem with `EventBus.clear` method. I believe it will affect entire event bus for all scopes when only single app stops.

PR #1011, opened 2019-06-08
Fix EventBus subscriptions leak, see #1009

carltonwhitehead:
The EventBus was holding references to subscriptions beyond the app lifecycle. This was a problem for my test automation, which was starting and stopping multiple `tornadofx.App` instances inside a single process. Stale subscriptions were getting called and completely wreaking havoc.

This PR makes two changes:
- Remove the EventBus subscriptions associated to a `Scope` when it is de-registered.
- Clear the EventBus entirely when an App stops.


edvin:
Nice work, thank you so much!

PR #1024, opened 2019-07-04
add ability to setup comparator to SortedFilteredList, added example

sith:
It looks like that SortedFilteredList doesn't have ability to setup comparator.

### DIFF
--- a/src/main/java/tornadofx/Lib.kt
+++ b/src/main/java/tornadofx/Lib.kt
@@ -38,9 +38,11 @@ class SortedFilteredList<T>(
         val items: ObservableList<T> = FXCollections.observableArrayList(),
         initialPredicate: (T) -> Boolean = { true },
         val filteredItems: FilteredList<T> = FilteredList(items, initialPredicate),
+        val comparatorProperty: ObservableValue<java.util.Comparator<T>>? = null,
         val sortedItems: SortedList<T> = SortedList(filteredItems)) : ObservableList<T> {
 
     init {
+        comparatorProperty?.let {  sortedItems.comparatorProperty().bind(it)}
         items.onChange { refilter() }
     }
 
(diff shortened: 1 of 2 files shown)


edvin:
I wonder if this approach would be more convenient to use:

```kotlin
class SortedFilteredList<T>(
        val items: ObservableList<T> = FXCollections.observableArrayList(),
        initialPredicate: (T) -> Boolean = { true },
        val filteredItems: FilteredList<T> = FilteredList(items, initialPredicate),
        val sortedItems: SortedList<T> = SortedList(filteredItems),
        comparator: java.util.Comparator<T>? = null) : ObservableList<T> {

    val comparatorProperty: ObjectProperty<java.util.Comparator<in T>>
        get() = sortedItems.comparatorProperty()
    var comparator by comparatorProperty

    init {
        comparator?.let { comparatorProperty.value = it }
        items.onChange { refilter() }
    }
...
}
```

First of all we won't break compatibility, and it's probably more realistic that you want to send in a Comparator instance, not a property. Now you can also bind a property if you want, the way you did it.

What do you think?

sith:
I am good with this approach.
Although I have 2 suggestions:
1. refilter function is doing sorting  so better name with javadoc would be helpful here. refilter function could be marked as deprecated and new one with better name is introduced. 
2. the main constructor of SortedFilteredList is a little bit messy. Now there will be 5 parameters and it is not clear which one should be set, and I think it is error prone. For example:
`private val numbers = SortedFilteredList(filteredItems = FilteredList<MyItem>(FXCollections.observableArrayList()))`
adding to `numbers` does nothing as it adds to underlying `items` collection that is not connected now. So I think constructor should have only 3 parameters, original source(with default value if needed), initialFilter and comparator

PR #1059, opened 2019-08-23
Harmonize finally processing for Task and FXTask.

### LINKED ISSUE #1058: FxTask does not expose properties for succeeded / failed / cancelled
Currently FxTask and by extension the TaskStatus object does not allow to handle/distinguish between success / failure / cancel. All these events are delegated only to the completed property / finally function.

Proposal is to add properties or handlers for the success / failure / cancel events.

Romanow88:
Get rid of runtime exception.

This fixes #1058


Romanow88:
seems that travis has an issue loading the oraclejdk8. Probably a good idea to use 'Build dist: trusty' instead of xenial

Romanow88:
Or use openjdk8

Issue #1069, opened 2019-09-07
SmartTableCell style reset


Warkst:
I feel like `styleClass.clear()` is missing from the `updateItem` implementation of [`SmartTableCell`] (and probably others, like list cell and tree cell):

https://github.com/edvin/tornadofx/blob/6032256dc72aeb7637f9ba53f1ac5ecf0418cfad/src/main/java/tornadofx/TableView.kt#L71

Right now, `style = null` is called, but `styleClass` is not being cleared, which can lead to empty cells still having certain style classes despite the underlying item being cleared. I've had to jump through quite some hoops to be able to have a cell cache and still clear the style classes...

edvin:
That makes sense. Do you have time to submit a PR? If not, I’ll fix it asap, in a few days :)

Warkst:
I couldn't figure out how to attach it here, so I created a new PR at #1070. This one is just for the table cells, but I'll make separate PRs for the others now.

PR #1070, opened 2019-09-08
Fix for #1069

Warkst:
`SmartTableCell` now resets `styleClass` when `updateItem` is called
with a `null` item, or when `empty = true`.


Warkst:
Actually I screwed this up, I'll create a new PR that doesn't use my master branch... hang on

PR #1073, opened 2019-09-08
Fix for #1069

Warkst:
`SmartTableCell` now resets `styleClass` when `updateItem` is called
with a `null` item, or when `empty = true`.

Goody, this feels better!

On a related note, I saw that you use a utility method `cleanUp` in `SmartTreeCell`, but not in the two others. I could clean that up, while I'm at it, if you want. Would you let me know in a reply to the PR?


edvin:
Yes, please do!

Warkst:
Done! Should I be worried about Travis?

edvin:
No, don't worry about travis :) Will fix or revert the travis config issues when I have time :)

PR #1082, opened 2019-10-01
Add clipping to view transitions

ruckustboom:
It's clipped by default, but can be turned off with a simple Boolean flag.

Note that if the user wants more complicated clipping, that is still possible by overriding the `setup()` function. This just sets a useful default.


ruckustboom:
I don't want to merge this in directly, as I'm having some issues getting TornadoFX to run properly for me on JDK11. I can run it, but I have to hack my way around the OSGi checks, so I'd rather someone who doesn't have to do that test it.

ruckustboom:
I need to close this to work on other tasks. I'll open a new PR properly.

Issue #760, opened 2018-07-24
Proposal: Remove snake case from CssElementDelegate


ruckustboom:
CSS Elements don't use snake case by default as they are usually the class name, so it doesn't really make sense to include the option in the delegate.

Removing it would leave two options

1. Use the property name literally as the default value
2. Always require the user to pass in a name

The first one would be more consistent with how the other rules are defined, but may be pointless as using the proper names may lead to the naming conflicts that kept us from including them in the first place.

ruckustboom:
As it stands now:

```
val rectangle by csselement()  // Becomes `rectangle`
val Rectangle by csselement()  // Becomes `rectangle` (val may conflict with `Rectangle` class)
val myRectangle by csselement()  // Becomes `my-rectangle`
val rectangle by csselement("Rectangle")  // Becomes `Rectangle`
```

With option 1:

```
val rectangle by csselement()  // Becomes `rectangle`
val Rectangle by csselement()  // Becomes `Rectangle` (val may conflict with `Rectangle` class)
val myRectangle by csselement()  // Becomes `myRectangle`
val rectangle by csselement("Rectangle")  // Becomes `Rectangle`
```

With option 2:

```
val rectangle by csselement()  // INVALID
val Rectangle by csselement()  // INVALID (val may conflict with `Rectangle` class)
val myRectangle by csselement()  // INVALID
val rectangle by csselement("Rectangle")  // Becomes `Rectangle`
```

edvin:
I believe option 2 will cause less confusion, and the fix for existing code is simple. Anybody else want to weigh in?

ruckustboom:
I guess there is also the option to just leave the snake case option, but always default it to false. That way the user can override it if they want:

```kotlin
val myRectangle by csselement()  // Becomes `myRectangle`
val myRectangle by csselement(snakeCase = true)  // Becomes `my-rectangle`
```

Granted, if we leave that in, I don't see any advantage over just leaving the current implementation, as it's at least consistent with the other `CSSRule`s.

ruckustboom:
I realize I never actually finished this task. Are you still thinking option 2 would be good Edvin? If so, I'd recommend the change for the jdk10 branch (and thus for TFX 2).

edvin:
[...]

PR #1086, opened 2019-10-02
Require explicit name for CSS Element selectors

ruckustboom:
Also cleanup when snake case calculation takes place.


ruckustboom:
Closes #760

PR #1090, opened 2019-10-09
Update JDK 10 With Changes From Master

ruckustboom:
I just did a simple merge of `master` into `jdk10` with some naive conflict resolution. There could be quite a bit that was not done correctly, but the tests all pass for me.

I'm not familiar with travis, so I have no idea if the configuration is still correct. Also, it may be worth verifying all the version in `pom.xml`.


edvin:
Thank you so much @ruckustboom! I think this is the best approach to get things moving, so let's see how it goes!

PR #1099, opened 2019-10-19
changed the pom to only use gpg sign when the deploy command of maven…

ronsmits:
somehow the gpg is always connected to the install phase in maven instead of deploy. So this one word fix is needed (again :) )

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -278,7 +278,7 @@
                 <executions>
                     <execution>
                         <id>sign-artifacts</id>
-                        <phase>install</phase>
+                        <phase>deploy</phase>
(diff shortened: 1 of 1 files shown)


edvin:
This doesn't actually work, signing happens too late for some reason, so we need it this way to do automatic deployments. However, it's better that I just change that when I do release deploys, so I'll merge this PR now.

ronsmits:
It solved the problem for me :)

edvin:
Yeah, it's better this way :) I'm aware of it, so I can easily adjust before deployment. Better than creating problems for everybody else!

PR #1104, opened 2019-10-21
Jdk10

SchweinchenFuntik:
`jdk 10` branch reduced to `master` branch state. Have questions about the `AutoCompleteComboBoxSckin` file, is it needed

PR #1105, opened 2019-10-22
Fix issued caused by divergent PRs

ruckustboom:
This re-adds new TFX2 features that were accidentally reverted in #1104.


edvin:
I'm sorry, @ruckustboom, I should have caught that. Thank you!

ruckustboom:
No problem at all. It was an easy mistake and an easy fix :)

PR #1108, opened 2019-10-26
little improvements for calling asynchronous operation on UIComponent

SchweinchenFuntik:
more convenient way to do asynchronous work for UIComponent


edvin:
Nice :)

PR #1126, opened 2019-12-06
updated the dokka version to 0.10.0

kzzt:
Updated dokka to latest version (0.10.0) in the pom file to get rid of the "package not found" error.


edvin:
Thanks!

PR #1129, opened 2019-12-08
Remove non-nullable bound from runAsyncWithProgress

### LINKED ISSUE #1128: runAsyncWithProgress with nullable return type
The `runAsyncWithProgress` function is currently (1.7.19) defined as:

```
fun <T : Any> Node.runAsyncWithProgress(progress: Node, op: () -> T): Task<T>
```

The `Any` bound on `T` prohibits me from using nullable types like `String?`. This is very inconvenient when the operation can have a 'null result' (by being cancelled for example).

I noticed that `runAsync` *does* support nullable values, is there a reason for this difference or is it some kind of oversight? I couldn't find anything in the source code that would cause issues with possibly-null values.

HoldYourWaffle:
Fixes #1128.

I don't have the time to set up a development environment right now so I just edited it in the github UI, meaning there could be very obvious and dumb mistakes/oversights. If tests complain I'll fix it up.

### DIFF
--- a/src/main/java/tornadofx/Async.kt
+++ b/src/main/java/tornadofx/Async.kt
@@ -238,7 +238,7 @@ fun Node.runAsyncWithProgress(latch: CountDownLatch, timeout: Duration? = null, 
  * The default progress node is a ProgressIndicator that fills the same
  * client area as the parent. You can swap the progress node for any Node you like.
  */
-fun <T : Any> Node.runAsyncWithProgress(progress: Node = ProgressIndicator(), op: () -> T): Task<T> {
+fun <T> Node.runAsyncWithProgress(progress: Node = ProgressIndicator(), op: () -> T): Task<T> {
     if (this is Labeled) {
         val oldGraphic = graphic
         (progress as? Region)?.setPrefSize(16.0, 16.0)


edvin:
Merged :) Will update 2.0 the same way.

HoldYourWaffle:
This madlad has enough faith in random first-time contributors to merge PR's without waiting for the tests 😮 

Is it possible for me to use this fix right away or do I have to wait until the full 1.17.20 release?

edvin:
Haha, I checked it locally and verified that it wouldn't break any of my projects in addition :) I'll push a new snapshot to oss.sonatype.org right now.

edvin:
Hm.. I got 401 when deploying to nexus. You'd have to pull and build it locally for now I guess. Will investigate tomorrow!

HoldYourWaffle:
I think you [succeeded](https://oss.sonatype.org/service/local/repositories/snapshots/content/no/tornado/tornadofx/1.7.20-SNAPSHOT/tornadofx-1.7.20-20191208.225113-95.pom), it was published a couple of minutes ago and fixes the issue.

By the way, thank you for your immensely quick response & fix! This really was a deal-breaking issue for me, but I was able to get back up and running within half an hour 😄

edvin:
Thanks :) Just remember, if you run into an issue with a function in the framework, you can always submit a PR, or even just fork that function in your own app if you need slightly different behavior. Most functions in TornadoFX are pretty short and easy to fork when needed!

PR #1130, opened 2019-12-09
Fix Travis CI

abhinayagarwal:
All current Travis CI jobs are failing with the same error:

> The job exceeded the maximum log length, and has been terminated.

In deploy step, `tests` run again causing the same output twice. Moreover, we do not have keys to sign artifacts in place. Therefore, this PR removes both these steps while deploying.


edvin:
Thank you!

PR #1135, opened 2019-12-17
Keep v2 up to date with master

ruckustboom:
Note that this also changes the JavaFX version from 11 to 13. This shouldn't be a problem, but if that's not desired I can revert it back to 11.


edvin:
You're the best, @ruckustboom :))

PR #1141, opened 2019-12-26
Add destructuring operators to XYChart.Data

HoldYourWaffle:
I added `component1` and `component2` destructuring operators to `XYChart.Data` to enable code like this:
```kt
maxFitnessesSeries.data.forEach { (x, y) ->
    // exciting stuff
}
```


edvin:
Nice :)

PR #1146, opened 2020-01-03
Add destructuring operators to MapChangeListener

HoldYourWaffle:
I added destructuring operators to `MapChangeListener` to allow usage like this:

```kt
observableMap.addListener(MapChangeListener { (key, added, removed) ->
	// exciting stuff
})
```

This might be more controversial than #1141 since the order is relatively arbitrary (although it feels quite natural to me).
If this addition is not desired feel free to close this pull request, otherwise I should probably also add these operators to the other collection types' listeners.

### DIFF
--- a/src/main/java/tornadofx/Collections.kt
+++ b/src/main/java/tornadofx/Collections.kt
@@ -575,6 +575,10 @@ class SetConversionListener<SourceType, TargetType>(targetList: MutableList<Targ
     }
 }
 
+operator fun <K, V> MapChangeListener.Change<K, V>.component1(): K = key;
+operator fun <K, V> MapChangeListener.Change<K, V>.component2(): V? = valueAdded;
+operator fun <K, V> MapChangeListener.Change<K, V>.component3(): V? = valueRemoved;
+
 fun <T> ObservableList<T>.invalidate() {
     if (isNotEmpty()) this[0] = this[0]
 }


ruckustboom:
I'm not sure I agree the order is so arbitrary. Pretty much all change listeners in JavaFX use the order: (observable, old, new), which would correspond to to (`map`, `valueRemoved`, `valueAdded`). You could then add `key` as a 4th component, and you might as well add `wasRemoved` and `wasAdded` as the 5th and 6th while you're at it.

Unfortunately that would make the example a little uglier:
```kotlin
observableMap.addListener(MapChangeListener { (_, removed, added, key) ->
	// exciting stuff
})
```
But it would be more in line with JavaFX std-lib.

ruckustboom:
Also, I'm not sure if it's still the case, but historically we would often not specify the return type when dealing with JavaFX collections and properties, and thus just pass the platform type up the line. It may be a little against the Kotlin way, but it means we aren't making assumptions for the user.

edvin:
I agree with @ruckustboom - the changelistener order should be maintained if we add this. Not sure of the value if we do though, as the example illustrates.

HoldYourWaffle:
Sorry for not responding for so long, but I just got back into TornadoFX and funnily stumbled into this again.

If I understand correctly there are two changes I should make:
1. Don't specify the return type explicitly but pass the platform type (non-idiomatic but fully justified IMHO)
2. Change the order of the components to `observable`, `old`, `new`, `key`, `wasRemoved`, `wasAdded` (ugly but more consistent)

Would that be correct?

PR #1154, opened 2020-01-22
Added initialDirectory to chooseFile

MGlolenstine:
After wanting to do so, I found out that that's not a feature, even tho it is for a chooseDirectory.
Opening a PR for an addition of that feature.

### DIFF
--- a/src/main/java/tornadofx/Dialogs.kt
+++ b/src/main/java/tornadofx/Dialogs.kt
@@ -70,10 +70,11 @@ enum class FileChooserMode { None, Single, Multi, Save }
  *
  * If the user cancels, the returnedfile list will be empty.
  */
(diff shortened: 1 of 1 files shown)


edvin:
This sounds reasonable, but I seem to remember there was a reason for it not being added. @ruckustboom - do you remember a discussion about this?

ruckustboom:
Sorry for the late reply, but unfortunately I don't have much to offer. I do remember having this discussion before, but for the life of me, I can't remember the details.

MGlolenstine:
I've tested it thoroughly and I haven't noticed any problems with it (so far), so I think that the changes do pass... But keep in mind, that I've only tested it with Arch Linux and KDE Plasma with Dolphin. So if anyone wants to test it with other platforms, feel free to do so.

edvin:
OK, let's give it a go. I don't see any issues with it either :)

PR #1168, opened 2020-02-18
add AppStopListener

Mr-Po:
I added a AppStopListener.
We can do something when the app stopping.

e.g：
src/test/kotlin/tornadofx/testapps/AppStopListenerTest.kt


edvin:
We already have shutdown hooks, just do:

```kotlin
        beforeShutdown { 
            // Your shutdown task here
        }
```

You can also override `stop()` in your `App` class.

PR #1183, opened 2020-03-14
Improving MapConversionListener

Thijsiez:
Hej Edvin, thank you for all your work on TornadoFX, I love it! But I also found what I think is a bug...

When using the `MapConversionListener` with a `TargetType` which doesn't implement `equals()`, it will not fire remove events to the bound list.
This seems to be because the converted object we want to remove is recreated but doesn't match the original since it's a different (recreated) object and does not implement `equals()`. This is the case with certain `javafx.scene.Node` objects like `Hyperlink` and `Button` when used with [`EventTarget.bindChildren`](https://github.com/edvin/tornadofx/blob/27ba7813b28900170073059b13f6583064d80770/src/main/java/tornadofx/FX.kt#L647), new items will be added but old ones aren't removed.
I took inspiration from the `SetConversionListener` as that implementation worked fine with both `Hyperlink` and `Button` nodes. Instead of using the entire original `Entry` to reference the converted object, we just use its key.

Hope this helps! :)

Cheers,
Thijs

### DIFF
--- a/src/main/java/tornadofx/Collections.kt
+++ b/src/main/java/tornadofx/Collections.kt
@@ -425,8 +425,8 @@ fun <SourceTypeKey, SourceTypeValue, TargetType> MutableList<TargetType>.bind(
     val listener = MapConversionListener(this, ignoringParentConverter)
     if (this is ObservableList<*>) {
         sourceMap.forEach { source ->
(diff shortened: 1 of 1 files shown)


edvin:
Nice, thanks!

Issue #1263, opened 2020-07-26
Support for creating components using constructor via dependency injection


TalosDx:
koin supports inject with a constructor, which is a good way to describe exactly what is needed without code bloat, especially when there are quite a few class properties

But tornadofx itself creates its components through an empty constructor, without using di container

init di container
`FX.dicontainer = object : DIContainer, KoinComponent {
    override fun <T : Any> getInstance(type: KClass<T>): T =
        getKoin().get(clazz = type, qualifier = null, parameters = null)
}`

example of view
`class LoginView(windowController: WindowController, windowView: WindowView) : View("Auth") {
    override val root = vbox {
        windowView.root
    }
}`

error
`SEVERE: Uncaught error
java.lang.InstantiationException: dev.talosdx.pentaquark.view.auth.LoginView
	at java.lang.Class.newInstance(Class.java:427)
	at tornadofx.FXKt.find(FX.kt:434)
	at tornadofx.FXKt.find$default(FX.kt:423)
	at tornadofx.App.start(App.kt:83)`

TalosDx:
I do not know all the subtleties, but I can suggest this option
FX.kt
```Kotlin
fun <T : Component> find(type: KClass<T>, scope: Scope = FX.defaultScope, params: Map<*, Any?>? = null): T {
...
val cmp = dicontainer?.getInstance(type) ?: type.java.newInstance()
...
}
```

SchweinchenFuntik:
I have code like this, instead of parameters in the constructor, you can use this:
```
class LineChart24Hours : Fragment() {
    val titleChart: String by param("") // default
    val titleAxisY: String by param("")
    val tickUnitStart: Double by param(100.0)
    val lowerBound: Double by param(0.0)
    val upperBound: Double by param(0.0)
    val isAutoRanging: Boolean by param(true)
    val isForceZeroInRange: Boolean by param(false)
}

class ViewTestChar() : View() {
    val chart: LineChart24Hours by fragment(
        params = mapOf(
            LineChart24Hours::upperBound.name to 10000.0,
            LineChart24Hours::titleAxisY.name to "Title"
        )
    )
```

also why don't you want to inject controllers?

TalosDx:
Firstly, this is validation, when the compiler checks (in the case of koin) everything that is needed before starting the application.
Second is the speed and ease of writing code.
Thirdly, I can change di without any problems, since 90% supports inject via the constructor
Fourth, making such changes should not create problems.
[...]

PR #1264, opened 2020-07-28
Support for creating components using constructor via dependency injection

TalosDx:
[Details here](https://github.com/edvin/tornadofx/issues/1263)
Sorry about my english

### DIFF
--- a/src/main/java/tornadofx/FX.kt
+++ b/src/main/java/tornadofx/FX.kt
@@ -18,6 +18,7 @@ import javafx.scene.layout.Pane
 import javafx.scene.layout.VBox
 import javafx.stage.Stage
 import javafx.stage.Window
+import tornadofx.FX.Companion.dicontainer
 import tornadofx.FX.Companion.inheritParamHolder
 import tornadofx.FX.Companion.inheritScopeHolder
 import tornadofx.FX.Companion.stylesheets
@@ -176,6 +177,7 @@ class FX {
                 object : SimpleObjectProperty<(Class<out Component>?) -> String>({ it?.name ?: "Messages" }) {
                     override fun invalidated() = loadMessages()
                 }
+
         /**
          * Provides the name of the Resource Bundle for a given Component Class.
          * A `null` value may be passed to represent the global bundle.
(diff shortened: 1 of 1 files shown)


SchweinchenFuntik:
Your option gives priority to non-native DI. It also becomes unclear where we got the instance from (from which DI).

TalosDx:
dicontainer property of the FX class in which the changes take place.
Yes, this gives priority to the non-native di if the native di is not registered as a dicontainer

SchweinchenFuntik:
for other DIs you need to use:
```
val MyView : View() {
    val helloService: HelloService by di()
}
```

What are the reasons for TornadofX DI behavior change?

TalosDx:
My idea is to make it possible to use di through the constructor, ideally it should be for the native di too.
If you have ideas better suggest

SchweinchenFuntik:
Yes, but this requires some serious work on the concept rather than changing a few lines (maybe it should be done in Tornadofx 2.0).

Where is your constructor with parameters here? You just get an instance from your DI scope, nothing has changed on the Tornadofx side.

Also, will your DI account for the difference between view and fragment?

TalosDx:
Lunch is over, I'll write it in the evening

edvin:
If we do a major rewrite of the dependency injection mechanism in TornadoFX, I suggest making third party injection frameworks a first class citizen perhaps? TornadoFX could ship with a simple mechanism like it has now, with the ability to plug in a third party extension for all the roles the native mechanism is used for.

SchweinchenFuntik:
Well, there is one option for automatic inject through the constructor (through the same reflection). And it shouldn't break old code.

The logic is simple, we check the parameters of the primary constructor and look for the necessary dependencies in the scope.

The problem can only cause recursive component dependencies

Today or tomorrow I will try to provide the code.

SchweinchenFuntik:
https://github.com/SchweinchenFuntik/tornadofx/blob/DI_constructor_injection/src/main/java/tornadofx/FX.kt#L437

simple sample: 
```
class TView(private val tController: TController, val personModel: PersonModel) : View("Constructor Injection") {
    override val root = vbox(10) {
        paddingAll = 5
        textfield(personModel.name)

        listview(tController.persons) { cellFormat { text = it.name } }
    }
}

class TController(val personModel: PersonModel) : Controller() {
    val persons = observableListOf(
            Person().apply { name = "Bob" },
            Person().apply { name = "Vanga" }
    )

    init {
        personModel.item = persons.first()
    }
}

class Person {
    val idProperty = SimpleIntegerProperty(0)
    var id by idProperty

    val nameProperty = SimpleStringProperty("")
    var name by nameProperty
}

class PersonModel : ItemViewModel<Person>() {
    val id = bind(Person::idProperty)
    val name = bind(Person::nameProperty)
}
```

PR #1283, opened 2020-10-04
Adding an option to make drawer with horizontal text resolve #1267

### LINKED ISSUE #1267: Change the rotation of drawer items
For the Drawer view, the items are automatically rotated if they're placed on either side, and I suppose it is done by this method:

https://github.com/edvin/tornadofx/blob/904617a6e2e1582f810eae1ee569c600fcc2498d/src/main/java/tornadofx/Drawer.kt#L153-L159

I would like these items to be put horizontally so it can be read more easily. I suppose that I can do that with CSS, but I don't know how. Here are the approaches I tried:

```kotlin
drawer {
    item("Item 1", expanded = true) {
        addClass(DrawerViewStyle.horizontalDrawerItem)  // no effect
        style { rotate = 0.deg } // no effect
    }
    style { rotate = 0.deg } // no effect
    addClass(DrawerViewStyle.horizontalDrawerItem)  // no effect
}
```

Stylesheet class:

```kotlin
class DrawerViewStyle: Stylesheet() {
    companion object {
        val horizontalDrawerItem by cssclass()
        val item by css class()
    }
    init {
        horizontalDrawerItem {
            rotate = 0.deg
        }
        button {  // no effect
            rotate = 0.deg
        }
        item {  // no effect
            rotate = 0.deg
        }
    }
}
```
I didn't find any guide on this and the document doesn't really help. Is there any simple way I can handle this?

Huy-Ngo:
This PR isn't done yet - I haven't written a Kotlin library before and moreover I'm not familiar with maven, so I don't know how to test and write test and was struggling to do so in the past month. Hopefully you can help me finishing this PR.

### DIFF
--- a/src/main/java/tornadofx/Drawer.kt
+++ b/src/main/java/tornadofx/Drawer.kt
@@ -23,7 +23,7 @@ fun EventTarget.drawer(
         op: Drawer.() -> Unit
 ) = Drawer(side, multiselect, floatingContent).attachTo(this, op)
 
-class Drawer(side: Side, multiselect: Boolean, floatingContent: Boolean) : BorderPane() {
+class Drawer(side: Side, multiselect: Boolean, floatingContent: Boolean, horizontalItem: Boolean) : BorderPane() {
     val dockingSideProperty: ObjectProperty<Side> = SimpleObjectProperty(side)
     var dockingSide by dockingSideProperty
 
@@ -151,10 +151,14 @@ class Drawer(side: Side, multiselect: Boolean, floatingContent: Boolean) : Borde
     }
 
     private fun configureRotation(button: ToggleButton) {
-        button.rotate = when (dockingSide) {
-            Side.LEFT -> -90.0
-            Side.RIGHT -> 90.0
-            else -> 0.0
+        button.rotate = if (this.horizontalItem) {
+            when (dockingSide) {
+                Side.LEFT -> -90.0
+                Side.RIGHT -> 90.0
+                else -> 0.0
+            }
+        } else {
+            0.0
         }
     }
 
--- a/src/main/java/tornadofx/Drawer.kt
+++ b/src/main/java/tornadofx/Drawer.kt
@@ -5,6 +5,7 @@ import javafx.beans.value.ObservableValue
 import javafx.collections.FXCollections
 import javafx.event.EventTarget
 import javafx.geometry.Orientation
+import javafx.geometry.Pos
 import javafx.geometry.Side
 import javafx.scene.Group
 import javafx.scene.Node
@@ -341,6 +342,7 @@ class DrawerStyles : Stylesheet() {
         val drawerItem by cssclass()
         val buttonArea by cssclass()
         val contentArea by cssclass()
+        val horizontalDrawerItem by cssclass()
     }
 
     init {
@@ -370,6 +372,10 @@ class DrawerStyles : Stylesheet() {
             content {
                 borderColor += box(Color.TRANSPARENT)
             }
+        }
+        horizontalDrawerItem {
+            minWidth = 200.px
+            alignment = Pos.CENTER_LEFT
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/main/java/tornadofx/Drawer.kt
```diff
@@ -321,7 +326,10 @@ class DrawerItem(val drawer: Drawer, title: ObservableValue<String?>? = null, ic
                 if (change.wasAdded()) {
                     change.addedSubList.asSequence()
                             .filter { VBox.getVgrow(it) == null }
```
SchweinchenFuntik, line 328: no check HBox.getHgrow(it) == null

src/main/java/tornadofx/Drawer.kt
```diff
@@ -23,7 +24,7 @@ fun EventTarget.drawer(
         op: Drawer.() -> Unit
 ) = Drawer(side, multiselect, floatingContent).attachTo(this, op)
```
SchweinchenFuntik, line 25: the old constructor is not used here

src/main/java/tornadofx/Drawer.kt
```diff
@@ -321,7 +326,10 @@ class DrawerItem(val drawer: Drawer, title: ObservableValue<String?>? = null, ic
                 if (change.wasAdded()) {
                     change.addedSubList.asSequence()
                             .filter { VBox.getVgrow(it) == null }
-                            .forEach { VBox.setVgrow(it, Priority.ALWAYS) }
+                            .forEach {
+                                VBox.setVgrow(it, Priority.ALWAYS)
```
SchweinchenFuntik, line 330: there is a potential increase in memory. You need to think about how to check which type of Panel is used

src/main/java/tornadofx/Drawer.kt
SchweinchenFuntik: append default value

src/main/java/tornadofx/Drawer.kt
```diff
@@ -23,7 +24,7 @@ fun EventTarget.drawer(
         op: Drawer.() -> Unit
```
SchweinchenFuntik, line 24: append parameter `horizontalItem`

src/main/java/tornadofx/Drawer.kt
```diff
@@ -23,7 +24,7 @@ fun EventTarget.drawer(
...
 ) = Drawer(side, multiselect, floatingContent).attachTo(this, op)
 
-class Drawer(side: Side, multiselect: Boolean, floatingContent: Boolean) : BorderPane() {
+class Drawer(side: Side, multiselect: Boolean, floatingContent: Boolean, horizontalItem: Boolean) : BorderPane() {
```
SchweinchenFuntik, line 27: it might be worth renaming. For example, by adding the prefix ʻis`, it is possible to remove ʻItem`.

SchweinchenFuntik (commented):
I have not yet understood what problem you are trying to solve, you have not referred to the issue

Huy-Ngo:
> you have not referred to the issue

I thought I already linked it by mentioning in the title, but for some reason GitHub didn't link it. I was trying to resolve #1267. I just edited my comment to show what I expect.

src/main/java/tornadofx/Drawer.kt
Huy-Ngo: Should I add a check for horizontal item here? In that case, I think I need to add a parameter for `DrawerItem` in order to check.

SchweinchenFuntik: might be worth redesigning ExpandedDrawerContentArea itself

cilki:
Update on this?

ThraaxSession:
@edvin any idea when this comes into the upstream? I am eager to use it. :)

Huy-Ngo:
Sorry, as I didn't have any experience writing a Kotlin library so I don't know how to make sure my patch works. Also, there are some code issues as Funtik pointed out.

SchweinchenFuntik:
@Huy-Ngo you can run tests directly in the IDE. Using task `test`

Issue #1318, opened 2021-03-16
Fragment instance leaks when using replaceWith()


mswokk:
I had a memory leaking problem when replacing Fragments.
And I found the cause and fixed it.

Cause: 
Leaking point is in Component.kt > rootSceneChangeListener
When adding a windowProperty() change listener, it uses onChangeOnce() utility method.
onChangeOnce() is designed to unregister the listener itself when it called once.
But replacing the scene with Fragment or VIew, the windowProperty is not changed and not called
Hence, the window instance holds the windowProperty() change listener and never releases.

Minimal reproduction:
Repeat several times to calling ...
replaceWith<AFragment>() from AView , 
replaceWith<AView>() from AFragment

And check the JVM heap histogram with jmap command.
jmap -histo:live <pid> | grep AFragment

The number of instance of AFragment should be one, but it wasn't.

I created a PR for this issue.

PR #1319, opened 2021-03-16
[#1318] Fix leaking fragments

mswokk:
Fix leaking Fragments.

Remove windowProperty() listener before undocking.

And some refactoring for listeners registering for dock/undock.
Move listener holders from properties (map) to backing field by lazy() kotlin to reduce indent and lines.

Even the fixes pass the test cases, I couldn't find a way to check memory leaking.
It should be verified manually using jmap histogram dump.

Please review the code.

### DIFF
--- a/src/main/java/tornadofx/Component.kt
+++ b/src/main/java/tornadofx/Component.kt
@@ -528,61 +528,53 @@ abstract class UIComponent(viewTitle: String? = "", icon: Node? = null) : Compon
         properties["tornadofx.closeable"] = SimpleBooleanProperty(false)
     }
 
-    private val rootParentChangeListener: ChangeListener<Parent>
-        get() {
-            val key = "tornadofx.rootParentChangeListener"
-            if (properties[key] == null) {
-                properties[key] = ChangeListener<Parent> { _, oldParent, newParent ->
-                    if (modalStage != null) return@ChangeListener
-                    if (newParent == null && oldParent != null && isDocked) callOnUndock()
-                    if (newParent != null && newParent != oldParent && !isDocked) {
-                        callOnDock()
-                        // Call `onTabSelected` if/when we are connected to a Tab and it's selected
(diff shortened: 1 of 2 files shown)


mswokk:
Thank you for review @SKeeneCode 

The rootParentChangeListener instance is still hold by the property `rootParentChangeListener`
I just moved the reference from `Component.properties` to property `rootParentChangeListener` explicitly.
So it doesn't break the bug fix for docking.

I couldn't consider the `rootParentChangeListener` is refered from outside of the class.
Because the key `tornadofx.rootParentChangeListener` was declared in the function locally.
If users need to use `rootParentChangeListener` to unregiser and register again manually, 
IMHO, the `rootParentChangeListener` should be declared with `open` access modifier.
It's better than refered by string key from a HashMap, where the string key is not accessible from outside. 
Declaring `rootParentChangeListener` as a property is more explicit and simple.

And I'm not sure about this refactoring is really breaking change for most of common users.

mswokk:
Thank you for review @SKeeneCode 

The rootParentChangeListener instance is still hold by the property `rootParentChangeListener`
I just moved the reference from `Component.properties` to property `rootParentChangeListener` explicitly.
So it doesn't break the bug fix for docking.

SKeeneCode:
I think this looks good - could you apply this bugfix to the [tornadofx2 ](https://github.com/edvin/tornadofx2) repo and then I can modify the [reference ](https://github.com/edvin/tornadofx2/blob/master/src/main/java/tornadofx/Workspace.kt#L315) in the workspace (or you could do it).

I actually think the way I went about fixing the bug was a poor way. I think instead of removing and re-adding the listener I can just do `muteDocking = false`. Which means a reference to the listener is never needed.

guizmaii:
Hi @SKeeneCode,

I've done what you're asking here: https://github.com/edvin/tornadofx2/pull/40

